PR #51, opened 2024-08-30
test: refactor test data generation

SebastianElvis:
This PR refactors the tests, including

- making [create_new_finality_provider](https://github.com/babylonlabs-io/babylon-contract/blob/c8f6ab64ac00a934bb1570d5188f2405da7a09dc/contracts/btc-staking/src/contract.rs#L401-L402) generate valid finality providers
- making [get_derived_btc_delegation](https://github.com/babylonlabs-io/babylon-contract/blob/3335ab7b6a3b82be241852eb8db6a523bf62b699/contracts/btc-staking/src/contract.rs#L351-L352) generate valid BTC delegations
- removing the babylon-private dependency

This is the prerequisite for testing the full validation of finality providers / BTC delegations.


datagen/utils/btcstaking.go
```diff
@@ -79,13 +155,31 @@ func GenBTCDelegation(dir string) {
...
-	paramsBytes, err := bsParams.Marshal()
-	require.NoError(t, err)
-	paramsPath := filepath.Join(dir, BTCSTAKING_PARAMS_FILENAME)
-	err = os.WriteFile(paramsPath, paramsBytes, 0644)
-	require.NoError(t, err)
+func GenBTCDelegationsAndParams(dir string) {
+	covenantSKs, covenantQuorum := GenParams(dir)
+	// 3 FPs
+	GenFinalityProviders(dir, 3)
+	// 2 BTC delegations under 1st FPs
```
maurolacy, line 175:
```suggestion
	// 3 BTC delegations under 1st FPs
```
?

datagen/utils/btcstaking.go
```diff
@@ -79,13 +155,31 @@ func GenBTCDelegation(dir string) {
...
+	GenBTCDelegations(dir, covenantSKs, covenantQuorum, 1, []int{1})
+	GenBTCDelegations(dir, covenantSKs, covenantQuorum, 2, []int{1})
+	GenBTCDelegations(dir, covenantSKs, covenantQuorum, 3, []int{1})
+	// 2 BTC delegations under 2nd FPs
```
maurolacy, line 179:
```suggestion
	// 3 BTC delegations under 2nd FPs
```
?

Can the BTC del IDs overlap across different FPs?

contracts/btc-staking/src/queries.rs
```diff
@@ -296,25 +296,29 @@ mod tests {
...
         let fp1 = FinalityProvider::from(&new_fp1);
         let fp2 = FinalityProvider::from(&new_fp2);
-        assert_eq!(fps.len(), 2);
-        assert_eq!(fps[0], fp1);
-        assert_eq!(fps[1], fp2);
+        let expected_fps = vec![fp1.clone(), fp2.clone()];
+        assert_eq!(fps.len(), expected_fps.len());
+        for fp in fps {
+            assert!(expected_fps.contains(&fp));
+        }
```
maurolacy, line 306: An alternative here and below is to first sort both arrays before comparing them.

maurolacy (approved):
LGTM. Just some issues with data files name formatting, and some outdated module refs that would be nice to fix before merging.

.gitmodules
```diff
@@ -1,8 +1,4 @@
...
-	path = packages/proto/babylon
-	url = git@github.com:babylonlabs-io/babylon.git
-	branch = dev
-[submodule "packages/proto/babylon-private"]
-	path = packages/proto/babylon-private
-	url = git@github.com:babylonlabs-io/babylon-private.git
-	branch = base/consumer-chain-support
+path = packages/proto/babylon
+url = git@github.com:babylonlabs-io/babylon.git
+branch = dev
```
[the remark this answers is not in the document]
maurolacy: Already fixed this myself, hope you don't mind.

datagen/utils/finality.go
```diff
@@ -49,8 +48,6 @@ func GenRandomPubRandList(r *rand.Rand, numPubRand uint64) (*datagen.RandListInf
 }
 
 func GenCommitPubRandListMsg(startHeight uint64, numPubRand uint64, pubRandIndex uint64, dir string) (*datagen.RandListInfo, *btcec.PrivateKey) {
-	r := rand.New(rand.NewSource(time.Now().UnixNano()))
-
```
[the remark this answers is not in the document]
SebastianElvis: Yes. This seems to be a good practice. The issue I was facing is as follows. I was using two fresh `r := rand.New(rand.NewSource(time.Now().UnixNano()))` to generate secret keys for covenant members and finality providers in test data. Since the code was fast, the two `r` end up using the same timestamp and being the same value. Then the 1st covenant secret key is same as the 1st finality provider secret key. As Babylon does not accept duplicate keys in the staking script (i.e., any duplication among FP keys, covenant keys and staker key), the tests panicked. Using the same rand source solves this issue.

datagen/utils/btcstaking.go
SebastianElvis: It won't. BTC delegation ID is the staking tx hash, and the staking tx is certainly unique

contracts/btc-staking/src/queries.rs
SebastianElvis: Tried, but somehow the test still failed, probably the results are returned out of order. That's why I ended up simply checking inclusion

packages/test-utils/testdata/btc_delegation_1_{1,3}.dat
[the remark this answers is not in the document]
SebastianElvis: These are the identifiers of the FPs that this BTC delegation restakes to

datagen/utils/finality.go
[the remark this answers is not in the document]
maurolacy: Makes sense overall, but I find difficult to believe that both `r` have the same seed, as it's a timestamp in nanoseconds resolution. Are these routines being called in parallel or so?

PR #58, opened 2024-09-09
chore: move validation functions to a new mod

SebastianElvis:


contracts/btc-staking/src/validation/mod.rs
```diff
@@ -75,3 +83,165 @@ pub fn verify_new_fp(new_fp: &NewFinalityProvider) -> Result<(), ContractError>
...
+/// It is marked with `#[cfg(not(feature = "full-validation"))]` so that it
+/// is not included in the build if the `full-validation` feature is enabled.
+#[cfg(not(feature = "full-validation"))]
+pub fn verify_active_delegation(
+    _params: &Params,
+    _active_delegation: &ActiveBtcDelegation,
+    _staking_tx: &Transaction,
+) -> Result<(), ContractError> {
+    Ok(())
+}
```
maurolacy, line 247: Nit: An alternative that is perhaps cleaner, is instead of having two functions, have just one and put the `full-vaildation` feature in the body.

The warning about missing parameters can be avoided by doing something like

`let _ = (params, active_delegation, staking_tx)` in an extra `not` section.

maurolacy (approved):
LGTM.

contracts/btc-staking/src/validation/mod.rs
SebastianElvis: good ideas!

PR #60, opened 2024-09-12
btcstaking: full validation of unbonding/slashing BTC delegation

SebastianElvis:
Part of #7 

This PR implements the full validation of unbonded/slashed BTC delegations.


contracts/btc-staking/src/validation/mod.rs
```diff
@@ -49,6 +50,42 @@ fn verify_pop(
     Ok(())
 }
 
+#[cfg(feature = "full-validation")]
+fn get_pks(
```
maurolacy, line 54: Nit: this helper should perhaps be better named `deserialize_pks`, `decode_pks` or so.

contracts/btc-staking/src/staking.rs
```diff
@@ -221,12 +223,18 @@ fn handle_undelegation(
     // Basic stateless checks
     undelegation.validate()?;
 
+    let params = PARAMS.load(storage)?;
+
```
maurolacy, line 227: Nit: This can be moved below, after line 235.

contracts/btc-staking/src/error.rs
```diff
@@ -50,6 +50,8 @@ pub enum ContractError {
     FinalityProviderNotFound(String),
     #[error("Staking tx hash already exists: {0}")]
     DelegationAlreadyExists(String),
+    #[error("BTC delegation is not active")]
+    DelegationIsNotActive,
```
maurolacy, line 54: For clarity:
```suggestion
    #[error("BTC delegation is not active: {0}")]
    DelegationIsNotActive(String),
```
and use the staking tx hash or so for reference.

contracts/btc-staking/src/validation/mod.rs
```diff
@@ -49,6 +50,42 @@ fn verify_pop(
...
+    staker_pk_hex: String,
+    fp_pk_hex_list: Vec<String>,
+    cov_pk_hex_list: Vec<String>,
```
maurolacy, line 57: Use references to avoid early cloning.
```suggestion
    staker_pk_hex: &str,
    fp_pk_hex_list: &[String],
    cov_pk_hex_list: &[String],
```

or even `&[&str]` if possible / valid.

contracts/btc-staking/src/validation/mod.rs
```diff
@@ -241,3 +255,107 @@ pub fn verify_active_delegation(
 
     Ok(())
 }
+
+pub fn verify_undelegation(
```
maurolacy, line 259: Not sure about this, but perhaps this can be a method of `BtcDelegation`. Problem is that that will require extending the feature flag to the package.

contracts/btc-staking/src/validation/mod.rs
```diff
@@ -241,3 +255,107 @@ pub fn verify_active_delegation(
...
+            unbonding_path_script.as_script(),
+            &staker_pk,
+            &staker_sig,
+        )
+        .map_err(|e| ContractError::SecP256K1Error(e.to_string()))?;
+    }
+
+    // make static analyser happy with unused parameters
+    #[cfg(not(feature = "full-validation"))]
+    let _ = (params, btc_del, sig);
```
maurolacy, line 318: This is ugly but effective.

An alternative could be, to decorate the call to the method with the `full-validation` feature flag. So that there are no extra `not` sections.
On the other side, the nice thing about this format, is that we could be consolidating basic and full validations in the same helper.

contracts/btc-staking/src/validation/mod.rs
```diff
@@ -241,3 +255,107 @@ pub fn verify_active_delegation(
...
+
+    Ok(())
+}
+
+pub fn verify_slashed_delegation(
+    active_delegation: &BtcDelegation,
+    slashed_fp_sk_hex: String,
```
maurolacy, line 325: Here as well, it's in general better to use 
```suggestion
    slashed_fp_sk_hex: &str,
```

contracts/btc-staking/src/validation/mod.rs
```diff
@@ -241,3 +255,107 @@ pub fn verify_active_delegation(
...
+        // calculate the corresponding VerifyingKey
+        let slashed_fp_pk = slashed_fp_sk.verifying_key();
+        let slashed_fp_pk_hex = hex::encode(slashed_fp_pk.to_bytes());
+
+        // check if the PK corresponds to a FP PK that the delegation restakes to
+        if !active_delegation
+            .fp_btc_pk_list
+            .contains(&slashed_fp_pk_hex)
+        {
+            return Err(ContractError::FinalityProviderNotRegistered);
```
maurolacy, line 352: Here as well, let's indicate which one.

maurolacy (approved):
LGTM. Just some style / syntax suggestions.

Nice that we have these complex validations in place, btw.

contracts/btc-staking/src/validation/mod.rs
SebastianElvis: The verification of an early unbonded BTC delegation is dfferent from verifying the undelegation info inside `BTCDelegation`. The former verifies whether the staker's signature over unbonding tx is valid. The latter verifies whether the staker provides correct unbonding tx, unbonding slashing tx, and signs relevant signatures correctly.

contracts/btc-staking/src/validation/mod.rs
SebastianElvis: Yep wrapping the entire feature inside this `validation/` package seems cleaner than having feature flag in the contract files.

PR #63, opened 2024-09-13
Disable recovered_fp_btc_sk validation in SlashedBtcDelegation

gusin13:
The Slashed Delegation IBC packet from Babylon -> Consumer expects to have FP SK, imo this is not needed in the contract side as it has nothing to do with the extracted key. 

For now i have disabled the validation which happens on packet receive, in later pr i'll need to fix the proto files both in babylon side and contract side. 

Ref https://github.com/babylonlabs-io/babylon/blob/7942dae93db7f61539e4f84ecc3ff67f32299f44/proto/babylon/btcstaking/v1/packet.proto#L136

### DIFF
--- a/packages/apis/src/validate.rs
+++ b/packages/apis/src/validate.rs
@@ -185,9 +185,9 @@ impl Validate for SlashedBtcDelegation {
             return Err(StakingApiError::InvalidStakingTxHash(HASH_SIZE * 2));
         }
 
-        if self.recovered_fp_btc_sk.is_empty() {
-            return Err(StakingApiError::EmptyBtcSk);
-        }
+        // if self.recovered_fp_btc_sk.is_empty() {
+        //     return Err(StakingApiError::EmptyBtcSk);
+        // }
 
         Ok(())
     }


SebastianElvis:
Isn't this needed for the consumer chain to verify the corresponding FP is indeed slashed?

maurolacy:
> Isn't this needed for the consumer chain to verify the corresponding FP is indeed slashed?

No if we trust Babylon.

packages/apis/src/validate.rs
```diff
@@ -185,9 +185,9 @@ impl Validate for SlashedBtcDelegation {
             return Err(StakingApiError::InvalidStakingTxHash(HASH_SIZE * 2));
         }
 
-        if self.recovered_fp_btc_sk.is_empty() {
-            return Err(StakingApiError::EmptyBtcSk);
-        }
+        // if self.recovered_fp_btc_sk.is_empty() {
+        //     return Err(StakingApiError::EmptyBtcSk);
+        // }
```
maurolacy, line 190: Nit: remove commented code.

maurolacy (approved):
If we remove this, which I'm OK with, let's just remove the commented code, and the field from the struct as well.

SebastianElvis:
Then how about we make this field optional, and then only deal with it when `full-validation` feature is enabled?

SebastianElvis:
> Why don't we simply send slashing evidence from Babylon, if the full-validation ff is turned on contract is responsible to extract the SK from evidence, this would avoid the extraction logic in Babylon (which in my understanding is bit slow)?

Yup this works as well, and it seems better as Babylon does less work :+1:

maurolacy:
> Why don't we simply send slashing evidence from Babylon, if the full-validation ff is turned on contract is responsible to extract the SK from evidence, this would avoid the extraction logic in Babylon (which in my understanding is bit slow)?

This is a good idea in principle. The only problem is, that at that point the evidence might not be available anymore, as this slashing messages comes from Babylon, after slashing propagation / cascading.

Issue #11, opened 2024-07-31
Consider having functionality separate into multiple contracts


maurolacy:
**[<img src="https://avatars.githubusercontent.com/u/11656534?s=17&v=4" width="17" height="17"> @maurolacy](https://www.github.com/maurolacy)** cloned issue [babylonchain/babylon-contract#131](https://www.github.com/babylonchain/babylon-contract/issues/131) on 2024-06-05: 

> Related to #130.
> 
> 
> 
> `btc-staking` is growing big, and is in fact containing functionality related to finality signatures, and the finality providers set.
> 
> 
> 
> Consider introducing a new `finality` contract to handle finality votes, power, and the voting round. The power that comes from`btc-staking` can be read through a raw query for performance.
> 
> 
> 
> See the `poe-contracts` (https://www.github.com/confio/poe-contracts) design / structure for reference / ideas.
> 
> 
> 
> Also, the voting round and the finality validator set, though related to finality, could be perhaps better be in its own `fpset` contract. Again, `poe-contracts` provides a reference for this.

SebastianElvis:
Subsumed by #52 and #53

PR #64, opened 2024-09-16
R/index btc height

maurolacy:
A small refactor that removes BTC height indexing, as the BTC height is only used during slashing, and there's no need to store it in the mean time.

Part of #11, having separate `btc-staking` and (a new) `babylon-finality` contracts.

### DIFF
--- a/contracts/btc-staking/src/contract.rs
+++ b/contracts/btc-staking/src/contract.rs
@@ -1,5 +1,3 @@
-use babylon_contract::msg::btc_header::BtcHeaderResponse;
-
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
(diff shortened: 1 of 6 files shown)


contracts/btc-staking/src/contract.rs
```diff
@@ -219,40 +214,13 @@ pub fn sudo(
 }
 
 fn handle_begin_block(deps: &mut DepsMut, env: Env) -> Result<Response<BabylonMsg>, ContractError> {
-    // Index BTC height at the current height
-    index_btc_height(deps, env.block.height)?;
-
```
SebastianElvis, line 224: This seems necessary for determining whether BTC delegations are expired due to timelock?

maurolacy: Yes, you're right. But we're currently gathering that info from Babylon. So, it's not needed at this point.

SebastianElvis: Have we decided to let Babylon side or consumer side to determine expiration of BTC delegations? I remember we are trying to figure out an algorithm that is executed by consumer side

maurolacy: Not sure. I think there's an issue for making Babylon forward those undelegations to the Consumers. Let me check.

maurolacy: I cannot find the issue, but I remember we discussed this at some point. @gusin13?

maurolacy: What about creating an issue to re-introduce this indexing at some point when it's needed?

SebastianElvis (approved):

contracts/btc-staking/src/contract.rs
maurolacy: Here: #66.

PR #65, opened 2024-09-16
R/finality contract

### LINKED ISSUE #52: refactor: split BTC staking contract and finality contract

maurolacy:
Part of #11. Closes #52.

TODO:

- [x] `babylon-finality` contract based on the `finality` module.
- [x] Refactor storage access to (smart) queries.
- [x] Refactor (batched) active finality provider set computation.
- [x] `babylon-finality` sudo handlers.
- [x] ~~(Cascaded)~~ `babylon-finality` instantiation logic.
- [x]  Config and Params handling.
- [x] Convert required unit test to multi-tests.
- [x] `full-validation` feature path.


contracts/btc-finality/src/state/config.rs
```diff
@@ -0,0 +1,36 @@
...
+pub(crate) const PARAMS: Item<Params> = Item::new("params");
+/// Storage for admin
+pub(crate) const ADMIN: Admin = Admin::new("admin");
+
+/// Config are Babylon-selectable BTC finality configuration
+// TODO: Add / enable config entries as needed
+#[cw_serde]
+pub struct Config {
+    pub denom: String,
+    pub babylon: Addr,
```
SebastianElvis, line 19: Does the finality contract needs to know babylon contract address? Probably not needed as Babylon's finality module does not need to talk to BTC light clients?

contracts/btc-finality/src/finality.rs
```diff
@@ -571,54 +655,7 @@ pub(crate) mod tests {
...
-            commitment: pub_rand.commitment.into(),
-            signature: pubrand_signature.into(),
-        };
-
-        let res = execute(deps.as_mut(), mock_env(), info.clone(), msg).unwrap();
-        assert_eq!(0, res.messages.len());
-    }
-
-    #[test]
+    #[ignore]
```
SebastianElvis, line 658: Why is this ignored? Same comments for other tests as well

SebastianElvis (approved):
Lgtm :+1:

contracts/btc-finality/src/state/config.rs
maurolacy: Yes, it does. Slashing is being sent from finality to babylon, for forwarding over IBC. In fact from there the slashing info is also propagated to the staking contract, for accounting (setting FP power to zero).

There's some coupling, but it's acceptable, and normal given the functionality.

Also, at the level of the instantiation logic, there's some back and forth to inform the contracts of the addresses of the other contracts. We can improve on that, by example by instantiating the contracts from the SDK. Will create an issue in `babylon-sdk` to do that in a follow-up.

contracts/btc-finality/src/finality.rs
maurolacy: Still WIP at the moment. Need to convert these ignored tests to the multi-test framework, as now there are multiple contracts involved as part of the test. Just one test standing, and this is good to go.

PR #70, opened 2024-09-25
R/test utils

maurolacy:
Refactor test utilities so that they are common to the contracts.

Also fixes the finality tests under the `full-validation` feature.


contracts/btc-finality/Cargo.toml
```diff
@@ -25,6 +25,8 @@ default = []
 cranelift = ["cosmwasm-vm/cranelift"]
 # for quicker tests, cargo test --lib
 library = []
+# feature for enabling the full validation
+full-validation = [ "btc-staking/full-validation" ]
```
SebastianElvis, line 29: Does the finality contract need the option for full validation? I thought the full validation is only relevant for Babylon <-> contract interaction, and only BTC staking contracts needs to have that. In theory finality contract can work with BTC staking contract with or without full validation.

For testing finality contract, we can always use BTC staking contract without full validation imo

maurolacy: Yes, but this is a stronger setting, in which confirm the `full-validation` feature works also in the multi-contract tests.

In fact all this comes from the fact that CI was running workspace tests with the full validation feature enabled, and they were failing in the finality contract.

SebastianElvis: I see. Is there a way to make this feature only for developments/tests? np with me if we cannot do it though

contracts/btc-staking/Cargo.toml
```diff
@@ -45,6 +45,7 @@ babylon-proto         = { path = "../../packages/proto" }
 babylon-btcstaking    = { path = "../../packages/btcstaking" }
 babylon-bitcoin       = { path = "../../packages/bitcoin" }
 eots                  = { path = "../../packages/eots" }
+test-utils       = { path = "../../packages/test-utils" }
```
SebastianElvis, line 48:
```suggestion
eots                  = { path = "../../packages/eots" }
test-utils            = { path = "../../packages/test-utils" }
```

SebastianElvis (approved):
:+1:

contracts/btc-finality/Cargo.toml
maurolacy: It's already like that. Because the code under the `library` feature, used when importing other contracts, is only used / invoked from test code. These high level `full-validation` features in contracts other than the staking contract are just wrappers to the staking one; so that the feature can be enabled on purpose. Which, only affects tests.

PR #71, opened 2024-09-26
fix: Fp info query should return err for non existent Fp's

gusin13:


SebastianElvis (approved):
Nice catch!

SebastianElvis:
Merging it to unblock the release of 0.9.0-rc.1

maurolacy (commented):
OK with me. We should strive to make these queries compatible / similar in behaviour to the Babylon ones, as much as possible.

PR #72, opened 2024-09-27
Fix/optimizer

maurolacy:
#65 small fix / follow-up.

Gates `library` feature included code (test code, basically), so that it's only included for tests and not the final wasm.
The need for this seems to arise from a limitation or particularity of the `test` config setting, which is not being propagated or set during inclusion of other code. Perhaps this is correct, as setting it could render that code unavailable (or different).

I don't know, more investigation would be needed to clarify / understand this better.

In the mean time, this is a (somewhat involved) way to solve / workaround this issue.


SebastianElvis (approved):
nice catch!

PR #73, opened 2024-09-27
F/optimizer ci

maurolacy:
Enables optimiser run during PRs / on branches.


SebastianElvis (approved):

maurolacy:
The only issue is that now the CI run goes from ~2minutes to ~10minutes :-/.

Perhaps we remove the cosmwasm check and integration tests? And we can still run those when merging to `main`.

SebastianElvis:
The check is still necessary imo we need to ensure 1) wasm binaries can indeed be executed on CosmWasm env, and 2) the contract size is not bloated. I don't have better idea for optimising it though

maurolacy:
This last version is a good in-between solution IMO. We check that contract binaries compile, and we run the full checks only when merging to `main` / on release tags.

maurolacy:
I would say, let's go with this, and we can tune it later if needed / considered.

maurolacy:
Also, we should consider building our own optimiser docker image and push it to docker hub. So that we don't build it every time. Will create an issue for it.

SebastianElvis:
> So that we don't build it every time. Will create an issue for it.

Good idea :+1: that shall slash the CI execution time a lot

PR #76, opened 2024-09-29
F/stock optimizer

maurolacy:
Replace building the optimiser by directly running the pre-built / stock optimiser.

This is using the `latest` tag of the optimiser image. It can be easily modified to use a fixed / manually defined version tag instead, for reproducibility / determinism. Also, we can force using the Intel (`amd64`) arch for the same goal / reason, instead of automatically switching to `arm64` / `amd64` images depending on the host's arch.


maurolacy:
Please notice that it still takes ~5 minutes to build the optimised binaries...

Running the checks, in which we need to build the `cosmwasm-check` from source first, takes even more. I think there's a way to use prebuilt binaries in Rust. Will take a look and eventually create an issue in the CosmWasm repo to suggest it.

bap2pecs (approved):

SebastianElvis:
:+1:

PR #80, opened 2024-10-30
F/rewards generation

maurolacy:
Call block reward generation on block finalisation.


maurolacy:
Closed in favour of #81.

PR #81, opened 2024-11-07
F/rewards generation 2

maurolacy:
A refactor of the rewards generation impl in #80.

Here we move all of the rewards calculation logic to the contract, which is good for integration and modularity.

It uses a `MintRewards` custom handler, which mints the rewards after validation / authentication of sender and recipient.

### DIFF
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -239,7 +239,10 @@ fn handle_end_block(
         let ev = finality::index_block(deps, env.block.height, &hex::decode(app_hash_hex)?)?;
         res = res.add_event(ev);
         // Tally all non-finalised blocks
-        let events = finality::tally_blocks(deps, activated_height, env.block.height)?;
+        let (msg, events) = finality::tally_blocks(deps, activated_height, env.block.height)?;
+        if let Some(msg) = msg {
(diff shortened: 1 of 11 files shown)


contracts/btc-finality/src/contract.rs
```diff
@@ -30,7 +30,34 @@ pub fn instantiate(
...
+        )?;
+        // Deserialize protobuf
+        let res_decoded = anybuf::Bufany::deserialize(&res).unwrap();
+        // See https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/proto/cosmos/mint/v1beta1/query.proto#L35-L36
+        let res_params = res_decoded.message(1).unwrap();
+        // See https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/proto/cosmos/mint/v1beta1/mint.proto#L60-L61
+        // to see from where the field number comes from
+        blocks_per_year = res_params
+            .uint64(6)
+            .ok_or(ContractError::MissingBlocksPerYear {})?;
```
SebastianElvis, line 56: Can we wrap this into a function?

SebastianElvis (approved):
Great work!

PR #85, opened 2024-11-19
Release 0.11

maurolacy:
Rewards generation reference release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,21 @@
 
 ## [Unreleased](https://github.com/babylonlabs-io/babylon-contract/tree/HEAD)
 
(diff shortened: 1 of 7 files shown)


SebastianElvis:
Wdyt we release after completing the entire reward user story (i.e., including sending inflation from consumer to Babylon)?

maurolacy:
Yes, that was my original idea. But I think it's good to have this released for reference anyway. Integrating with the new version of the contracts is still pending, so adding this delta makes sense IMO.

It's that, or we stick to releasing euphrates-0.6.0 without rewards altogether (i.e. v. 0.10.0 of the contracts). Let's discuss in the meeting today.

SebastianElvis (approved):

PR #86, opened 2024-11-20
Ignore now unsupported integration test

maurolacy:
Small #85 follow-up.

### DIFF
--- a/contracts/btc-finality/tests/integration.rs
+++ b/contracts/btc-finality/tests/integration.rs
@@ -21,6 +21,7 @@ fn wasm_size_limit_check() {
 }
 
 #[test]
+#[ignore = "Unsupported query type: GRPC"]
 fn instantiate_works() {
     let mut deps = mock_instance(WASM, &[]);
 


contracts/btc-finality/tests/integration.rs
```diff
@@ -21,6 +21,7 @@ fn wasm_size_limit_check() {
 }
 
 #[test]
+#[ignore = "Unsupported query type: GRPC"]
```
SebastianElvis, line 24: could you give more context about this one?

SebastianElvis (approved):

contracts/btc-finality/tests/integration.rs
maurolacy: The mocked environment in which integration tests are run don't support these queries. Since this query is now part of the finality contract instantiation, the instantiation integration test fails.

There's no easy way around this, except disabling or removing the test, as these integration tests are run against the optimised wasm binary that is going to be released.

PR #87, opened 2024-11-27
fix: bump contract and permissioned integration

SebastianElvis:
Accompanying with https://github.com/babylonlabs-io/babylon/pull/291

This PR

- bumps contract against the latest Babylon base branch
- fixes all failed tests due to the bump
- implements the consumer side logic of permissioned integration, mostly removing auto consumer registration


packages/btcstaking/src/tx_verify.rs
```diff
@@ -69,8 +69,7 @@ fn validate_slashing_tx(
     }
 
     // Verify that the first output pays to the provided slashing address.
-    let slashing_pk_script = slashing_address.script_pubkey();
-    if slashing_tx.output[0].script_pubkey != slashing_pk_script {
+    if slashing_tx.output[0].script_pubkey.as_bytes() != slashing_pk_script {
         return Err(Error::InvalidSlashingAddress {});
```
maurolacy, line 73: Nit:
```suggestion
        return Err(Error::InvalidSlashingScript {});
```
?

scripts/protocgen.sh
```diff
@@ -7,7 +7,7 @@ set -eo pipefail
 
 cd packages/proto
 # To also initialize, fetch and checkout any nested submodules
-git submodule update --init --recursive
+# git submodule update --init --recursive
```
maurolacy, line 10: Why commented?

maurolacy (commented):
Quick review of the current changes. Great work pulling this out.

scripts/protocgen.sh
SebastianElvis: Good catch, was for saving time of local tests. Fixed now

packages/proto/src/gen/babylon.btcstaking.v1.rs
```diff
@@ -345,56 +345,56 @@ pub struct NewFinalityProvider {
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ActiveBtcDelegation {
     /// staker_addr is the address to receive rewards from BTC delegation.
-    #[prost(string, tag="11")]
+    #[prost(string, tag="1")]
```
maurolacy, line 348: A lot of trouble caused by this. These tags are ids, and as such, they don't need to be sequential, start from 1, have a delta of 1, etc. They just have to be valid integers, and be unique in a given message.

For backwards compatibility reasons (and ease of integration), it makes much more sense to keep these tags unchanged. I know it looks "bad" or disorganised when these numbers are not sequential, but the truth is that keeping them unchanged wouldl make our lives easier.

Just mentioning this for awareness, perhaps at some point we can agree on this.

maurolacy (approved):
LGTM. Will review associated PRs next.

packages/proto/src/gen/babylon.btcstaking.v1.rs
SebastianElvis: In general I agree with this. Since phase-3 is still not live yet, we have the freedom of changing their values. For now we can try to make those clean. After the phase-3 testnet those values will not be changed unless there is really compelling reasons

PR #88, opened 2024-11-27
feat: add whitelist for reorg handling

parketh:
## Summary

This PR adds whitelisting for reorg handling. 

The current FG cannot handle L2 block reorgs because the new block hash breaks the finality votes query. This causes the finalized head to become stuck. 

We implement a new whitelist system in the CW contract to allow:
- a privileged admin to add whitelisted blocks
- FG to check blocks against the whitelist and skip them to unblock the derivation pipeline

## Test plan

```
cargo build
cargo optimize
cargo test
```

### DIFF
--- a/contracts/op-finality-gadget/src/exec/finality.rs
+++ b/contracts/op-finality-gadget/src/exec/finality.rs
@@ -315,6 +315,11 @@ pub fn whitelist_forked_blocks(
     // Check caller is admin
     check_admin(&deps, info)?;
 
+    // Check array is non-empty
+    if forked_blocks.is_empty() {
+        return Err(ContractError::EmptyBlockRange);
+    }
+
     // Append blocks to whitelist
     FORKED_BLOCKS.update::<_, ContractError>(deps.storage, |mut blocks| {
         blocks.extend(forked_blocks);
--- a/contracts/op-finality-gadget/src/exec/finality.rs
+++ b/contracts/op-finality-gadget/src/exec/finality.rs
@@ -3,7 +3,7 @@ use std::collections::HashSet;
 use crate::error::ContractError;
 use crate::queries::query_last_pub_rand_commit;
 use crate::state::config::CONFIG;
-use crate::state::finality::{BLOCK_VOTES, SIGNATURES};
+use crate::state::finality::{BLOCK_VOTES, FORKED_BLOCKS, SIGNATURES};
 use crate::state::public_randomness::{
     get_pub_rand_commit_for_height, PUB_RAND_COMMITS, PUB_RAND_VALUES,
 };
@@ -11,10 +11,12 @@ use crate::utils::query_finality_provider;
 
 use babylon_apis::finality_api::PubRandCommit;
 use babylon_merkle::Proof;
-use cosmwasm_std::{Deps, DepsMut, Env, Event, Response};
+use cosmwasm_std::{Deps, DepsMut, Env, Event, MessageInfo, Response};
 use k256::ecdsa::signature::Verifier;
 use k256::schnorr::{Signature, VerifyingKey};
 use k256::sha2::{Digest, Sha256};
+
+use super::admin::check_admin;
 
 // Most logic copied from contracts/btc-staking/src/finality.rs
 pub fn handle_public_randomness_commit(
@@ -301,6 +303,24 @@ pub(crate) fn verify_finality_signature(
         return Err(ContractError::FailedSignatureVerification("EOTS".into()));
     }
     Ok(())
+}
+
+/// `whitelist_forked_blocks` adds a set of forked blocks to the whitelist. These blocks will be skipped by FG
+/// (ie. treated as finalized) duringconsecutive quorum checks to unblock the OP derivation pipeline.
+pub fn whitelist_forked_blocks(
+    deps: DepsMut,
+    info: MessageInfo,
+    forked_blocks: Vec<(u64, u64)>,
+) -> Result<Response, ContractError> {
+    // Check caller is admin
+    check_admin(&deps, info)?;
+
+    // Append blocks to whitelist
+    FORKED_BLOCKS.update::<_, ContractError>(deps.storage, |mut blocks| {
+        blocks.extend(forked_blocks);
+        Ok(blocks)
+    })?;
+    Ok(Response::new())
 }
 
 /// `msg_to_sign` returns the message for an EOTS signature.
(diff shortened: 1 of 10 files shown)


contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -303,6 +305,29 @@ pub(crate) fn verify_finality_signature(
     Ok(())
 }
 
+/// `whitelist_forked_blocks` adds a set of forked blocks to the whitelist. These blocks will be skipped by FG
+/// (ie. treated as finalized) duringconsecutive quorum checks to unblock the OP derivation pipeline.
```
bap2pecs, line 309: during consecutive

contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -303,6 +305,29 @@ pub(crate) fn verify_finality_signature(
...
+pub fn whitelist_forked_blocks(
+    deps: DepsMut,
+    info: MessageInfo,
+    forked_blocks: Vec<(u64, u64)>,
+) -> Result<Response, ContractError> {
+    // Check caller is admin
+    check_admin(&deps, info)?;
+
+    // Check array is non-empty
+    if forked_blocks.is_empty() {
```
bap2pecs, line 319: should we also check the first element is less or equal to the second element in the tuple?

contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -303,6 +305,29 @@ pub(crate) fn verify_finality_signature(
...
+        return Err(ContractError::EmptyBlockRange);
+    }
+
+    // Append blocks to whitelist
+    FORKED_BLOCKS.update::<_, ContractError>(deps.storage, |mut blocks| {
+        blocks.extend(forked_blocks);
+        Ok(blocks)
+    })?;
```
bap2pecs, line 327: do we want to allow overalp?

e.g. `[(1,10) (5, 8) (3, 20)]`

contracts/op-finality-gadget/src/state/finality.rs
```diff
@@ -1,8 +1,11 @@
...
 use std::collections::HashSet;
 
 /// Map of signatures by block height and fp
 pub(crate) const SIGNATURES: Map<(u64, &str), Vec<u8>> = Map::new("fp_sigs");
 
 /// Map of (block height, block hash) tuples to the list of fps that voted for this combination
 pub(crate) const BLOCK_VOTES: Map<(u64, &[u8]), HashSet<String>> = Map::new("block_hashes");
+
+/// Ordered list of forked blocks [(start_height_1, end_height_1), (start_height_2, end_height_2), ...]
+pub(crate) const FORKED_BLOCKS: Item<Vec<(u64, u64)>> = Item::new("forked_blocks");
```
SebastianElvis, line 11: This structure means that the vector of start/end heights is considered as a single item. If we track a lot of forks (which might be the case), then the DB ops over it could be some bottleneck.

Would we consider other structures like `Map<u64, u64>` where key is the start height and value is the end height?

contracts/op-finality-gadget/src/exec/finality.rs
SebastianElvis: Agreed, would suggest we do not allow overlap. This simplifies some CRUD around the forked blocks

SebastianElvis (commented):
Great work!

contracts/op-finality-gadget/src/state/finality.rs
bap2pecs: > then the DB ops over it could be some bottleneck.

i think if it's ordered with no overlap, in practice it's gonna be efficient b/c we will loop in a reverse order. and it's unlikely the FG will need to read a lot of the old data to find the one its looking for

bap2pecs: If we store it as a map, i don't know how to design an efficient algorithm to decide if a block is reorged or not

contracts/op-finality-gadget/src/queries.rs
```diff
@@ -46,6 +47,51 @@ pub fn query_last_pub_rand_commit(
...
+    let forked_blocks = FORKED_BLOCKS.load(deps.storage)?;
+    for (start, end) in forked_blocks.iter().rev() {
+        // if block is in range of any forked block, return true
+        if height >= *start && height <= *end {
+            return Ok(true);
+        }
+        // terminate early once we reach a forked block range lower than the queried block
+        if height < *start {
+            break;
+        }
```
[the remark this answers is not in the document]
bap2pecs: actually we can just do a binary search

image

contracts/op-finality-gadget/src/state/finality.rs
SebastianElvis: I think `prefix_range` or other iterator functions can be useful here. Check out `prefixed_range_works` under cw-storage-plus` repo

bap2pecs: i just took a look. i think the mental model will be cleaner to simply use (u64, u64) and query with a naive binary search like this

image

SebastianElvis: Actually there is a much simpler way: given a height, use `prefix_range` to find the first key that is smaller than the height, and then check if the height is between the key and value

bap2pecs:
ok we discussed offline. so the issue was unlikely caused by an L1 reorg. instead, it's likely caused by the L1 RPC rate limiting thus causing the L2 reorgs

so this can still be useful but probably not a high-pri for now b/c usually L1 reorg depth is just 1 block and it has to get very unlucky to have a batch tx hit the exact L1 reorg block

SebastianElvis:
Btw, what do you guys think of having some test cases around this, to ensure that honest FPs won't sign two conflicting blocks due to l1 reorg. This seems important for the staker's money. If yes we can create github issues to track it

SebastianElvis:
> > Btw, what do you guys think of having some test cases around this, to ensure that honest FPs won't sign two conflicting blocks due to l1 reorg. This seems important for the staker's money. If yes we can create github issues to track it
> 
> iirc that's already guaranteed in the FP codebase.
> 
> actually the issue related to this PR already implies there is no double vote --- that's why FG stuck because it didn't find the signatures for reorged blocks

I see, then that means the recent addition of anti-slashing mechanisms work :+1:

SebastianElvis:
this work is now migrated to https://github.com/babylonlabs-io/rollup-bsn-contracts now

PR #89, opened 2024-11-28
Rm auto register 2

maurolacy:
A small follow-up to #87.

`cargo gen-proto` is failing btw. @SebastianElvis can you take a look?


SebastianElvis:
From what I see the submodule points to the `main` branch but not `base/consumer-chain-support` branch. The `main` branch does not contain zone concierge module and that's why `gen-proto` failed

SebastianElvis:
Created a small commit to fix this.

Btw, how about we use git clone to get the babylon repo, and then generate rust proto files? The submodule seems to create a lot of troubles

maurolacy:
Hmm, I like the submodule approach better.

It requires some administration though. Here are the steps that I know and use:

1. Add the submodule (`git submodule add`).
3. Select the specific branch / tag you want the submodule to point, using `git checkout` **inside** the submodule's checkout out repository.
4. Commit that change in the main repo.
5. You can now use `git update` to "checkout" that committed branch / tag / commit id in the submodule whenever you need to.
6. `git update --remote` updates the submodule ref to the latest commit in the specified branch. Not very useful, but still.

maurolacy:
I've now committed a change based in the sequence above.

The trick here is always committing the submodule's branch / commit changes into the main repo **before** running `cargo gen-proto`.
This is because we have `git submodule update` as part of the script, btw. We may decide to remove that, which will allow us to generate the proto files from a dirty / uncommitted submodule state.

maurolacy:
Now, after changing / updating the proto files, the `apis` package is broken. Something related to `delegator_unbonding_info` vs. `delegator_unbonding_sig`:

```rust
error[E0609]: no field `delegator_unbonding_info` on type `babylon_proto::babylon::btcstaking::v1::BtcUndelegationInfo`
   --> packages/apis/src/btc_staking_api.rs:246:14
    |
246 |             .delegator_unbonding_info
    |              ^^^^^^^^^^^^^^^^^^^^^^^^ unknown field
    |
help: a field with a similar name exists
    |
246 |             .delegator_unbonding_sig
    |              ~~~~~~~~~~~~~~~~~~~~~~~
```

maurolacy:
Yes, it works now.

SebastianElvis:
Hmm thanks for finding this out. We can change to use base branch after mering the permissioned integrtaion PR

SebastianElvis (approved):
Lgtm!

PR #92, opened 2024-12-03
feat: op consumer chain slashing

parketh:
## Summary

This PR implements slashing for op stack consumer chain.

It updates the cw contract as follows:

- allow FPs to equivocate by submitting mutiple signatures for a block
- add new state to index voted block hashes and equivocation evidences
- add new handlers to slash finality providers and send notifying message to Babylon,  similar to the Cosmos consumer chain implementation

It also updates and add a new test case. We need to add additional integration tests to check the e2e flow with Babylon node.

## Test plan
```bash
# build the contract
cargo build
cargo optimize

# test it
cargo test
cargo integration
```


contracts/op-finality-gadget/src/ibc.rs
SebastianElvis, line 1: Hmm why would this smart contract support IBC?

Babylon contract needs IBC because it is deployed on consumer chains and needs to establish IBC channels with Babylon. The finality gadget contract does not have this requirement

contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -328,13 +308,99 @@ fn check_fp_exist(deps: Deps, fp_pubkey_hex: &str) -> Result<(), ContractError>
...
+    let mut res = Response::new();
+    // Send over IBC to the Provider (Babylon)
+    let channel = IBC_CHANNEL.load(deps.storage)?;
+    let ibc_msg = ibc_packet::slashing_msg(&env, &channel, evidence)?;
+    // Send packet only if we are IBC enabled
+    // TODO: send in test code when multi-test can handle it
+    #[cfg(not(any(test, feature = "library")))]
+    {
+        res = res.add_message(ibc_msg);
+    }
```
SebastianElvis, line 375: The slashing message is not sent via an IBC channel; rather it's by using a message in `res`. This message is sent from the contract to the Cosmos SDK level of Babylon. Babylon side needs a handler for handling this message. 

So in order to complete the slashing user story, we will need to develop the following

- In Babylon node, adding a new message handler that upon the evidence, slash the given FP
- In the contract, getting the evidence of an equivocating FP, and sending it to the above message handler

@maurolacy Do you have any example of CosmWasm-Cosmos SDK communication in hand?

contracts/op-finality-gadget/src/ibc.rs
bap2pecs: yeah sorry I didn't provide the context. our op contract should be much simpler since we don't need to have most of those logic

contracts/op-finality-gadget/src/state/config.rs
```diff
@@ -11,4 +12,7 @@ pub const IS_ENABLED: Item<bool> = Item::new("is_enabled");
 #[cw_serde]
 pub struct Config {
     pub consumer_id: String,
+    pub consumer_name: Option<String>,
+    pub consumer_description: Option<String>,
```
lesterli, line 16: these two fields are unnecessary, we can query them with `consumer_id` from Babylon if need

contracts/op-finality-gadget/src/state/config.rs
```diff
@@ -11,4 +12,7 @@ pub const IS_ENABLED: Item<bool> = Item::new("is_enabled");
...
+    pub babylon: Addr,
```
lesterli, line 17: what is `babylon` field? it is better to add a comment

contracts/op-finality-gadget/src/exec/finality.rs
parketh: ah got it, so because the contract is hosted on Babylon, we cant directly message the node rather than use IBC which is more for cross-chain comms with Cosmos consumers. some code examples would def be helpful 🙏

for the new message handler on Babylon, do you guys have any preference on the name and route of this new handler fn? I can start a PR to add it. 

also, do you think we need to amend the `BTCSlasher` program, or we can simply emit the equivocation evidence inside an `MsgAddFinalitySig` as for babylon fps?

SebastianElvis: > some code examples would def be helpful 

Hmm actually you can search for `add_message` or `add_submessage` in the code base. Both functions are used for sending a message back to the chain that hosts the smart contract (in this context Babylon). One concrete example is [here](https://github.com/babylonlabs-io/babylon-contract/blob/5f5ba8de0f04fc807a2c5683ab49235c8514cfc9/contracts/babylon/src/ibc.rs#L178-L197)

> for the new message handler on Babylon, do you guys have any preference on the name and route of this new handler fn? I can start a PR to add it.

Thanks, wdyt we create a message for handling equivocation evidences under `x/btcstaking`? Say something like 

```protobuf
rpc EquivocationEvidence(MsgEquivocationEvidence) returns (MsgEquivocationEvidenceResponse);
```

And then upon this message, we verify the evidence, and then call [SlashConsumerFinalityProvider](https://github.com/babylonlabs-io/babylon/blob/4898d8ddd62cb01b4908ef9291a85a11d6e37149/x/btcstaking/keeper/finality_providers.go#L122-L123) to slash this FP.

> also, do you think we need to amend the BTCSlasher program, or we can simply emit the equivocation evidence inside an MsgAddFinalitySig as for babylon fps?

This is probably not needed, although we might want to move the logic [here](https://github.com/babylonlabs-io/babylon/blob/4898d8ddd62cb01b4908ef9291a85a11d6e37149/x/finality/keeper/msg_server.go#L310-L319) to `SlashConsumerFinalityProvider`, such that `SlashConsumerFinalityProvider` will propagate slashing events to relevant consumers, and emit slashing events for the BTC staking tracker.

bap2pecs: > This is probably not needed

why is that? I thought we will these two steps:

- emit `EventSlashedFinalityProvider` in the new msg handler `EquivocationEvidence`
- modify BTCSlasher to subscribe to this msg and filter the event

SebastianElvis: If the event is still in type `EventSlashedFinalityProvider`, then BTC slasher will recognise it just like slashed Babylon FP, right?

bap2pecs: > If the event is still in type `EventSlashedFinalityProvider`, then BTC slasher will recognise it just like slashed Babylon FP, right?

from our discussion in https://www.notion.so/snapchain/OP-Stack-CC-Slashing-Spec-145a62189ff480cbb2bace2403e3c33c:

"The BTCSlasher program subscribes to `EventSlashedFinalityProviderevent` inside the [MsgAddFinalitySig](https://github.com/babylonlabs-io/vigilante/blob/main/btcstaking-tracker/btcslasher/slasher.go#L25) msgs from Babylon node." But this msg is only for Babylon.

but my understanding is we will add a new msg for `EquivocationEvidence`. so it won't be detected by BTCSlasher until we modify it.

am I missing anything?

bap2pecs: @SebastianElvis we also see there is a `x/btcstkconsumer` module. is it a better place for slashing consumer FPs or do you think `x/btcstaking` is still a better choice?

contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -328,13 +307,84 @@ fn check_fp_exist(deps: Deps, fp_pubkey_hex: &str) -> Result<(), ContractError>
...
+    let pk = eots::PublicKey::from_hex(fp_btc_pk_hex)?;
+    let btc_sk = pk
+        .extract_secret_key(
+            &evidence.pub_rand,
+            &evidence.canonical_app_hash,
+            &evidence.canonical_finality_sig,
+            &evidence.fork_app_hash,
+            &evidence.fork_finality_sig,
+        )
+        .map_err(|err| ContractError::SecretKeyExtractionError(err.to_string()))?;
```
SebastianElvis, line 326: Extracting BTC SK does not seem necessary here

contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -328,13 +307,84 @@ fn check_fp_exist(deps: Deps, fp_pubkey_hex: &str) -> Result<(), ContractError>
...
+    let msg = ExecuteMsg::Slashing {
+        evidence: evidence.clone(),
+    };
+    let wasm_msg: WasmMsg = WasmMsg::Execute {
+        contract_addr: env.contract.address.to_string(),
+        msg: to_json_binary(&msg)?,
+        funds: vec![],
+    };
+
+    let ev = Event::new("slashed_finality_provider")
```
SebastianElvis, line 339: Do we need to emit an event given that Babylon will do this instead?

SebastianElvis (approved):
Lgtm after some minor comments and merging the PR on Babylon side :+1:

contracts/op-finality-gadget/src/exec/finality.rs
parketh: i think its probably still useful to emit an event here. could be for debugging or building a slashing monitor in future.

parketh:
merging this after resolving comments, we can reopen with new PR if there are any other issues (e.g. after implementing some e2e tests)

contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -328,13 +307,84 @@ fn check_fp_exist(deps: Deps, fp_pubkey_hex: &str) -> Result<(), ContractError>
     }
 }
 
+/// `slash_finality_provider` slashes a finality provider with the given evidence including setting
+/// its voting power to zero, extracting its BTC SK, and emitting an event
```
SebastianElvis, line 311: WE do not extract SK here

contracts/op-finality-gadget/src/exec/finality.rs
```diff
@@ -328,13 +307,84 @@ fn check_fp_exist(deps: Deps, fp_pubkey_hex: &str) -> Result<(), ContractError>
...
+    Ok((wasm_msg, ev))
+}
+
+pub(crate) fn handle_slashing(evidence: &Evidence) -> Result<Response<BabylonMsg>, ContractError> {
+    let mut res = Response::new();
+    // Send msg to Babylon
+
+    let msg = BabylonMsg::EquivocationEvidence {
+        evidence: Some(evidence.clone()),
+    };
```
SebastianElvis, line 366: Probably the syntax needs to be fixed as per https://github.com/babylonlabs-io/babylon/pull/327

SebastianElvis (commented):
Lgtm apart from some minor stuff

contracts/op-finality-gadget/src/exec/finality.rs
parketh: sorry what do you mean by this?

PR #93, opened 2024-12-04
chore: use query_grpc from cosmwasm_std

lesterli:
## Summary

This small PR replaces our custom `query_grpc` method, which is no longer needed, as the upgraded `cosmwasm-std` (v2.1.4) now provides a built-in `query_grpc` method.　

## Test Plan

```
cargo lint
cargo unit-test
cargo integration
```

### DIFF
--- a/contracts/op-finality-gadget/src/utils.rs
+++ b/contracts/op-finality-gadget/src/utils.rs
@@ -1,8 +1,5 @@
 use anybuf::{Anybuf, Bufany};
-use cosmwasm_std::{
-    to_json_vec, Binary, ContractResult, Deps, GrpcQuery, QueryRequest, StdError, StdResult,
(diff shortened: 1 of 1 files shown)


bap2pecs (approved):

contracts/op-finality-gadget/src/utils.rs
```diff
@@ -53,29 +49,3 @@ pub fn query_finality_provider(
 
     Ok(res)
 }
-
-/// TODO: query_grpc need to be replaced with cosmwasm::std::QuerierWrapper.query_grpc
```
bap2pecs, line 57: to provide context, this is not needed anymore b/c the dependency has been upgraded

SebastianElvis (approved):
Nice :+1:

PR #94, opened 2024-12-06
refactor instantiation logic

SebastianElvis:
- [x] remove contract addr in config
- [x] remove cascaded instantiation logic
- [x] use custom query to get contract addresses
- [ ] unit tests
- [ ] e2e tests in Babylon SDK https://github.com/babylonlabs-io/babylon-sdk/pull/77


packages/bindings/src/babylon_sdk.rs
```diff
@@ -0,0 +1,53 @@
...
+            },
+        }
+    }
+}
+
+pub fn get_babylon_sdk_params(querier: &QuerierWrapper) -> Result<Params, StdError> {
+    let params = querier.query_grpc(QUERY_PARAMS_PATH.to_owned(), Binary::new("".into()))?;
+    let params = QueryParamsResponse::from(params).params;
+    Ok(params)
+}
```
maurolacy, line 53: I think this approach is not without merit, but it would be much simpler IMO to just pass the needed params during instantiation. The contracts have a clear dependencies, so it's just a matter of instantiating them in the right order and pass required parameters (other contract's addresses) between them at instantiation.

contracts/btc-finality/src/state/config.rs
```diff
@@ -17,8 +17,6 @@ pub(crate) const ADMIN: Admin = Admin::new("admin");
 pub struct Config {
     pub denom: String,
     pub blocks_per_year: u64,
-    pub babylon: Addr,
-    pub staking: Addr,
```
maurolacy, line 21: I would just keep these, and pass / set them during instantiation.

contracts/btc-finality/src/contract.rs
```diff
@@ -254,10 +260,12 @@ fn handle_end_block(
     Ok(res)
 }
 
-pub fn get_activated_height(staking_addr: &Addr, querier: &QuerierWrapper) -> StdResult<u64> {
+pub fn get_activated_height(querier: &QuerierWrapper) -> StdResult<u64> {
+    let params = get_babylon_sdk_params(querier)?;
```
maurolacy, line 264: Nit: This is good for encapsulation, but bad for performance. In any case, since you already have it why not pass the entire `Config` by reference instead?

maurolacy (approved):
LGTM. Just some comments about the overall design.

Will review the SDK PR next.

packages/bindings/src/babylon_sdk.rs
SebastianElvis: Finally got back to this PR. In the ideal case, indeed it's better if we let each contract to only keep references of contracts that it interacts with. However, we have a couple of cyclic dependencies between contracts, making it difficult for instantiating them in order.

In addition, the contract might not be persistent. It's possible that the consumer's gov prop deploys another sets of contracts and specifies them as the "official version" through the governance proposal. Once this happens then there can be discrepencies between parameters of Babylon SDK and Babylon contracts.

Thus I proposed we make contract addresses in Babylon SDK module parameters as single source of truth, and let contract find each other through Babylon SDK. What do you think?

PR #95, opened 2024-12-10
F/rewards distribution

maurolacy:
Implements first level (across FPs) rewards distribution.

This is an initial / naïve impl of rewards distribution. It will likely need to be extended / augmented to support better distribution for late finality votes, and / or depending on the active set instead of the voting table.

### DIFF
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -91,8 +91,8 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
         QueryMsg::Config {} => Ok(to_json_binary(&queries::config(deps)?)?),
         QueryMsg::Params {} => Ok(to_json_binary(&queries::params(deps)?)?),
         QueryMsg::Admin {} => to_json_binary(&ADMIN.query_admin(deps)?).map_err(Into::into),
-        QueryMsg::FinalitySignature { btc_pk_hex, height } => Ok(to_json_binary(
-            &queries::finality_signature(deps, btc_pk_hex, height)?,
+        QueryMsg::FinalityVote { btc_pk_hex, height } => Ok(to_json_binary(
+            &queries::finality_vote(deps, btc_pk_hex, height)?,
         )?),
         QueryMsg::PubRandCommit {
             btc_pk_hex,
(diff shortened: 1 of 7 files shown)


contracts/btc-finality/src/finality.rs
```diff
@@ -622,3 +627,35 @@ pub fn list_fps_by_power(
...
+        .range(deps.storage, None, None, Ascending)
+        .collect::<StdResult<Vec<_>>>()?;
+    // Get the total voting power of the voting FPS
+    let total_voting_power = voting_fps
+        .iter()
+        .map(|(_, vote)| vote.voting_power as u128)
+        .sum::<u128>();
+    // Get the rewards to distribute (bank balance of the staking contract)
+    let cfg = CONFIG.load(deps.storage)?;
+    let rewards_amount = deps.querier.query_balance(cfg.staking, cfg.denom)?.amount;
```
SebastianElvis, line 644: Since the protocol generates all rewards upon finalising blocks, would the following situation happen:

- Suddenly the protocol finalises 10 blocks and generates rewards for these blocks
- The protocol distributes the rewards of these 10 blocks to the FPs *in the last block*
- The FPs for the rest 9 blocks do not get reward

Babylon resolves this problem by having a [KV store](https://github.com/babylonlabs-io/babylon/blob/6f77345f0242765e5de3c5df261a431ebafe56ae/x/incentive/keeper/btc_staking_gauge.go#L79-L88) to track the reward for every height, and the protocol distributes rewards in each height's gauge to the FPs at that height. Maybe we need something similar?

contracts/btc-finality/src/contract.rs
```diff
@@ -91,8 +92,8 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
         QueryMsg::Config {} => Ok(to_json_binary(&queries::config(deps)?)?),
         QueryMsg::Params {} => Ok(to_json_binary(&queries::params(deps)?)?),
         QueryMsg::Admin {} => to_json_binary(&ADMIN.query_admin(deps)?).map_err(Into::into),
-        QueryMsg::FinalitySignature { btc_pk_hex, height } => Ok(to_json_binary(
-            &queries::finality_signature(deps, btc_pk_hex, height)?,
+        QueryMsg::FinalityVote { btc_pk_hex, height } => Ok(to_json_binary(
```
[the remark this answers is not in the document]
maurolacy: Yes, this is reverted now.

contracts/btc-finality/src/finality.rs
maurolacy: I tried to avoid doing that. See #97 description for rationale.

maurolacy:
Closing in favour of #97.

PR #96, opened 2024-12-10
F/rewards distribution send

maurolacy:
#97 follow-up. Sends the rewards along with the distribution table to Babylon using ICS-20 with a json payload in the Memo field.


contracts/babylon/src/contract.rs
```diff
@@ -269,6 +271,44 @@ pub fn execute(
...
+            let btc_finality = cfg
+                .btc_finality
+                .ok_or(ContractError::BtcFinalityNotSet {})?;
+            if info.sender != btc_finality {
+                return Err(ContractError::Unauthorized {});
+            }
+            // Build the memo payload
+            let memo_msg = to_json_string(&fp_distribution)?;
+            // Route to babylon over IBC
+            let channel = IBC_CHANNEL.load(deps.storage)?;
```
SebastianElvis, line 288: Hmm I guess `IBC_CHANNEL` is the channel that this contract involves, but not necessarily a channel for IBC transfer. Probably we need to confirm this in some e2e test.

Maybe as a part of the integration, one needs to establish an IBC channel for interchain transfer, and the contract side needs to remember this channel ID in the config, for facilitating the reward mechanism.

maurolacy: Yes, this was my suspicion as well. Did it way for simplicity. Perhaps we can configure a single channel to do / support both, custom messages and ICS20 transfers?

Will write an e2e test and confirm this in any case.

maurolacy: Finally was able to e2e test this.

Using a single channel doesn't work, as CosmWasm's `IbcMsg::Transfer` requires an established channel against the `ibctransfer` module, not the contract. :-/ I think this is a design decision that makes sense, as you don't want to establish a channel for every contract that needs / wants to use ICS20 transfers.

SebastianElvis: > Using a single channel doesn't work, as CosmWasm's IbcMsg::Transfer requires an established channel against the ibctransfer module, not the contract address under wasm.

Yes that's right, even though we can wire up a IBC transfer packet in the contract side, babylon side's zone concerige won't recognise it.

If we really, really want to let zone concerige handle IBC tansfer, then zone concierge has to invoke IBC transfer module (in particular [this](https://github.com/cosmos/ibc-go/blob/91169d134d659b87acaf4903652cdcb1a55f5657/modules/apps/transfer/ibc_module.go#L193-L204) code snippet) to handle the IBC transfer packets. But I'm not sure if it's worth it

So I'd recommend we consider consolidating IBC channels as future work, and for now start with having 2 IBC channels. This also creates a clean separation between channels' functionalities following IBC channels' best practices

contracts/btc-finality/src/finality.rs
```diff
@@ -632,14 +632,20 @@ pub fn distribute_rewards(deps: &mut DepsMut, env: &Env) -> Result<(), ContractE
     };
     // Get the voting power of the active FPS
     let total_voting_power = active_fps.iter().map(|fp| fp.power as u128).sum::<u128>();
-    // Get the rewards to distribute (bank balance of the finality contract)
+    // Get the rewards to distribute (bank balance of the finality contract minus already distributed rewards)
+    let distributed_rewards = TOTAL_REWARDS.load(deps.storage)?;
```
SebastianElvis, line 636: Is it possible to get the total reward by querying the contract's balance, rather than tracking it in a Kv store?

SebastianElvis (approved):

contracts/btc-finality/src/finality.rs
maurolacy: The problem with this is, that `TOTAL_REWARDS` are the rounded sum of the distributed rewards over finality providers. They are not necessarily the same. With this impl, we left the remainder in the balance, and it'll be distributed later eventually. Better than send an amount that doesn't match the sum of the distributed rewards, IMO.

maurolacy:
Merging as approved. Please take a look at recent changes (IBC transfer channel info) and comment back in any case.

TODOs (custom rewards distribution protocol) will be done in follow-up PRs.

PR #97, opened 2024-12-11
F/rewards distribution 2

maurolacy:
Follow-up / Alternative to #95, in which the rewards are distributed over the (latest) active finality provider set.

This may incur a number of (small) errors in the distribution, namely:

1. It rewards members of the active set wether they voted for a block or not. This is good for simplicity, as tracking the votes would require delaying rewards distribution until all the votes for a block are gathered. This is also what is being done on the Babylon side AFAIK.
2. It rewards members of the **latest** active set, even when some previous blocks might have been finalised over / with a different active set. In this impl we distribute pending rewards as soon as we have an active set available. Since the active set changes infrequently and (typically) slightly, this is good for simplicity, as in this way we don't need to track which rewards belong to which active set / height. This design decision is also supported by the fact that block rewards are small, and so reward errors per block, if any, will be accordingly small as well.
3. We consider the entire balance of the **finality** contract to be rewards, and we distribute it entirely, to the first available active finality provider set. Again, this is for simplicity, as in this way we don't need to track balance amounts at each block.
4. The remainder of the distribution across a given active fp set is left in the balance for further distribution. Since we're taking the total balance instead of specific balance amounts, we're preventing in this way a secular accumulation of remainders in the balance.

In summary: These small potential errors are considered acceptable, and are all in the name of simplicity.

Please notice that this also fixes a bug in the impl, in which we were considering the rewards were being sent to the staking contract for distribution. They clearly have to be sent to the **finality** contract instead, since a) it has the info for  distribution, b) it has to send the rewards to Babylon, and so it requires ownership of them.
Will fix this in the SDK and docs next as well.

### DIFF
--- a/contracts/btc-finality/src/finality.rs
+++ b/contracts/btc-finality/src/finality.rs
@@ -1,7 +1,8 @@
 use crate::contract::encode_smart_query;
 use crate::error::ContractError;
 use crate::state::config::{Config, CONFIG, PARAMS};
-use crate::state::finality::{BLOCKS, EVIDENCES, FP_SET, NEXT_HEIGHT, SIGNATURES};
+use crate::state::finality;
+use crate::state::finality::{BLOCKS, EVIDENCES, FP_SET, NEXT_HEIGHT, VOTES};
 use crate::state::public_randomness::{
     get_last_pub_rand_commit, get_pub_rand_commit_for_height, PUB_RAND_COMMITS, PUB_RAND_VALUES,
 };
@@ -192,9 +193,9 @@ pub fn handle_finality_signature(
         return Err(ContractError::HeightTooHigh);
     }
     // Ensure the finality provider has not cast the same vote yet
-    let existing_sig = SIGNATURES.may_load(deps.storage, (height, fp_btc_pk_hex))?;
-    match existing_sig {
-        Some(existing_sig) if existing_sig == signature => {
+    let existing_vote = VOTES.may_load(deps.storage, (height, fp_btc_pk_hex))?;
+    match existing_vote {
+        Some(existing_vote) if existing_vote.signature == signature => {
             deps.api.debug(&format!("Received duplicated finality vote. Height: {height}, Finality Provider: {fp_btc_pk_hex}"));
             // Exactly the same vote already exists, return success to the provider
             return Ok(Response::new());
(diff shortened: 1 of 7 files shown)


SebastianElvis:
Probably this design is good for now as 1) this serves as a PoC/baseline on the user stories that we can improve upon, and 2)  we haven't received a concrete tokenomics design from the product team. 

However, I think problem 1/2/3 needs to be resolved if we really decides on the pro rata distribution over FPs. This is because the protocol needs to ensure the reward distribution is fair, i.e., the reward of an FP has to be proportional to its contribution. Using the last voting power table to do the distribution might cause non-negligible bias. For example, if a FP is slashed and then suddenly the chain has 100 new finalised blocks, then this FP won't get any reward for its previous contribution. This looks like an edge case in terms of engineering, but this bias might not be negligible from the product's pov.

It would be good if we create a github issue to track this

contracts/btc-finality/src/finality.rs
```diff
@@ -622,3 +620,38 @@ pub fn list_fps_by_power(
...
+
+/// `distribute_rewards` distributes rewards to finality providers who are in the active set at `height`
+pub fn distribute_rewards(deps: &mut DepsMut, env: &Env) -> Result<(), ContractError> {
+    // Try to use the finality provider set at the previous height
+    let active_fps = FP_SET.may_load(deps.storage, env.block.height - 1)?;
+    // Short-circuit if there are no active finality providers
+    let active_fps = match active_fps {
+        Some(active_fps) => active_fps,
+        None => return Ok(()),
+    };
```
SebastianElvis, line 632: Maybe add a TODO here that we will have to use the voting power table of the height that is being finalised, rather than the latest height, for fair distribution.

Btw, Babylon side has implemented the exact design for reward distribution at the FP level -- https://github.com/babylonlabs-io/babylon/blob/main/x/incentive/keeper/btc_staking_gauge.go

SebastianElvis (approved):

maurolacy:
> For example, if a FP is slashed and then suddenly the chain has 100 new finalised blocks, then this FP won't get any reward for its previous contribution. This looks like an edge case in terms of engineering, but this bias might not be negligible from the product's pov.

If a FP falls off the active set it will lose their contribution from previous blocks, yes. It is assumed the delta of finalised blocks is typically / usually small (<10?).
On the other hand, if a FP is in the active set and doesn't vote, it gets rewards anyway (as long as blocks are being finalised by other FPs, that is). So, this serves as a kind of compensation.

> 
> It would be good if we create a github issue to track this

Will do.

PR #100, opened 2024-12-16
F/`btc-lc` feature

maurolacy:
Introduce a `btc-lc` feature flag, to remove BTC light client functionality from the Babylon contract.

`babylon-contract` has reached the wasm size limit (800KB), and to continue introducing features we must either move functionality to another contract, or reduce its size by removing features.

Here we do the second approach, which can in any case serve as a step towards the first one.


SebastianElvis:
Since we need to split BTC light client into another contract (https://github.com/babylonlabs-io/babylon-contract/issues/53), wdyt we go straight to that? Maintaining such a new feature would be very challenging and at the end of the day we will be removing this feature.

maurolacy:
May be. I did this to continue working on the rewards distribution feature, as the size limit was blocking me.

I would say let's merge this to unblock stuff, and work on the `btc-lc` contract ASAP. Another option could be to leave this on a branch. But, rewards distribution requires this, so not sure about it.

maurolacy:
In any case, we can merge this and then revert the entire squashed commit when working on the `btc-lc` contract.

SebastianElvis:
> I would say let's merge this to unblock stuff, and work on the btc-lc contract ASAP.

To unblock can we temporarily enlarge the size limit of wasm contract? This seems cleaner

maurolacy:
It seems, but I'm afraid it would involve changing the wasmd source code, i.e. we cannot directly use the binary artefact anymore.

I'll take a look.

SebastianElvis:
> It seems, but I'm afraid it would involve changing the wasmd source code, i.e. we cannot directly use the binary artefact anymore.
> 
> I'll take a look.

Not really, we did this once in Babylon SDK actually by setting some public variables.

maurolacy:
You're right, I was confused between wasmd and wasmvm. Thanks.

PR #104, opened 2025-01-21
F/rewards distribution consumer

### LINKED ISSUE #102: Native rewards distribution protocol
Alternatively to sending the rewards to Babylon using ICS-020 (#96), we can keep the rewards on the Consumer, for simplicity and a cleaner / simpler user experience.

~At the same time, we can still make Babylon to process the distribution table across delegators; for ownership / control, commissions, etc~. (Or we can do all the processing on the Consumer).

For this, we can establish a custom IBC message to send the distribution table across FPs to Babylon (instead of sending it on the Memo field of the ICS-020 transfer). And then another custom messages in the opposite direction, to send back the distribution table across delegators to the FPs. (Or send the distribution table and funds to the `btc-staking` contract, and process / distribute everything locally).

There's WIP for this already in the https://github.com/babylonlabs-io/babylon/tree/f/consumer-reward-distribution branch. Also, the babylon-contract `InstantiationMsg` has the optional  `transfer_info` field (https://github.com/babylonlabs-io/babylon-contract/blob/9fdc9a15de1f082bc0a0686fe4be6bf320a03a53/contracts/babylon/src/msg/contract.rs#L58) which can be used to disable ICS-020 entirely.

maurolacy:
Closes #102, with the variant in which we process all the distribution on the Consumer.

Implements the rewards distribution's second step (over delegators) on the Consumer.
Closely follows the mesh security impl, adapting it to our use case.

**TODO**:

- [x] Queries.
- [x] Unit tests.
- [x] Improvements / refactoring.
- [x] Renaming / review changes.


contracts/btc-staking/src/state/config.rs
```diff
@@ -49,4 +49,5 @@ pub struct Params {
     /// expressed as a decimal (e.g. 0.5 for 50%).
     #[derivative(Default(value = "String::from(\"0.1\")"))]
     pub slashing_rate: String,
+    pub rewards_denom: String,
```
SebastianElvis, line 52: Can this be queried from Cosmos SDK or be retrieved [here](https://github.com/babylonlabs-io/babylon-contract/blob/50fb1d70bfbbf3a24e0e60b59ab8cf40429d623f/contracts/babylon/src/state/config.rs#L26-L27)?

contracts/btc-staking/src/staking.rs
```diff
@@ -188,12 +196,43 @@ pub fn handle_active_delegation(
...
         // Update aggregated voting power by FP
-        fps.update(storage, fp_btc_pk_hex, height, |fp_state| {
-            let mut fp_state = fp_state.unwrap_or_default();
-            fp_state.power = fp_state.power.saturating_add(active_delegation.total_sat);
-            Ok::<_, ContractError>(fp_state)
-        })?;
+        fp_state.power = fp_state.power.saturating_add(active_delegation.total_sat);
+
+        // Create delegation distribution info. Fail if it already exists
+        delegations().delegation.update(
```
SebastianElvis, line 205: Could we wrap L205-232 to a function to keep this file cleaner?

contracts/btc-staking/src/staking.rs
```diff
@@ -288,17 +350,31 @@ fn handle_slashed_delegation(
...
-            let mut fp_state =
-                fp_state.ok_or(ContractError::FinalityProviderNotFound(fp.clone()))?; // should never happen
-            fp_state.power = fp_state.power.saturating_sub(btc_del.total_sat);
-            Ok::<_, ContractError>(fp_state)
-        })?;
+        let mut fp_state = fps.load(storage, &fp)?;
+        fp_state.power = fp_state.power.saturating_sub(btc_del.total_sat);
+
+
+        // Distribution alignment
```
SebastianElvis, line 357: Same, is it possible to define various functions for handling point alignments / reward per stake things

contracts/btc-staking/src/staking.rs
```diff
@@ -322,6 +398,164 @@ pub fn handle_slash_fp(
...
+                &mut deps,
+                &reward_info.fp_pubkey_hex,
+                reward_info.reward,
+                env.block.height,
+            )
+        })
+        .collect()
+}
+
+fn distribute_rewards(
```
SebastianElvis, line 433: Does this function distributes commission to FP's address, or this happens somewhere else?

contracts/btc-staking/src/state/delegations.rs
```diff
@@ -0,0 +1,102 @@
...
+/// pair, including distribution alignment
+#[cw_serde]
+pub struct Delegation {
+    /// The delegator's canonical address
+    pub staker_addr: CanonicalAddr,
+    /// How many satoshis the user stakes in this delegation
+    pub stake: u64,
+    /// How many points should be added / subtracted from points calculated per delegation due to
+    /// delegation changes.
+    pub points_alignment: PointsAlignment,
```
SebastianElvis, line 16: Not sure if I fully under this definition -- 

1. What is the definiton of "point" in this context?
2. Why would "the amount of points added/substracted" be a property of a delegation that is not affected by external factors?

contracts/btc-staking/src/state/delegations.rs
```diff
@@ -0,0 +1,102 @@
+use crate::state::points_alignment::PointsAlignment;
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{CanonicalAddr, Order, StdResult, Storage, Uint128};
+use cw_storage_plus::{Index, IndexList, IndexedMap, KeyDeserialize, MultiIndex};
+
+/// Single delegation related information - entry per `(staking hash, finality provider public key)`
+/// pair, including distribution alignment
+#[cw_serde]
+pub struct Delegation {
```
SebastianElvis, line 9: This is a bit confusing with the `BtcDelegation` struct. If this struct is dedicated for the reward distribution shall we rename it to something explicitly?

contracts/btc-staking/src/state/delegations.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+#[allow(clippy::type_complexity)]
+pub struct DelegationIndexes<'a> {
+    // Delegations by finality provider's public key and staking hash.
+    // Last type param defines the pk deserialization type
+    pub rev: MultiIndex<'a, (String, Vec<u8>), Delegation, (Vec<u8>, String)>,
+    // Delegations by staker's (raw, canonical) address and finality provider's public key.
+    // Last type param defines the pk deserialization type
+    pub staker: MultiIndex<'a, (Vec<u8>, String), Delegation, (Vec<u8>, String)>,
```
SebastianElvis, line 28: Not sure if I understand these two multi index:

- Why would the key contain both raw and canonical addresses? Probably raw address is good enough as the reward distribution does not care about the account prefix at all?
- What is the definition of `rev`?
- Why should we use multi index and differentiate key / primary key?

contracts/btc-staking/src/state/points_alignment.rs
SebastianElvis, line 1: I'm not entirely sure if this abstraction is needed or not. In BTC staking, for each BTC delegation with `x` BTC stake, the amount is either `x` or zero, but never in between. This `PointsAlignment` seems to track the points w.r.t. variable amount of stake, which does not seem necessary

SebastianElvis (commented):
Some initial comments, still parsing the whole thing

contracts/btc-staking/src/state/points_alignment.rs
maurolacy: Agreed. Points alignment make sense when you have multiple delegations aggregated by staker, and here we're splitting the distribution info over delegations already, so it's is not needed.

This is closely following the mesh security implementation. We can decide to aggregate distribution info 
by staker, for simplicity, and because it makes sense, that's why I've decided to keep this.
My idea is to test the impl, and then refactor if worth it for simplicity and clarity.

Thanks for the review. Will address your other comments later.

contracts/btc-staking/src/state/config.rs
maurolacy: Fixed. Queried using the `query_bonded_denom`, which makes sense for consistency with other parts of the code.

That said, rewards denom and bonded denom are not necessarily the same. This is something we will need to address at some point.

contracts/btc-staking/src/state/delegations.rs
maurolacy: Yes, we can rename to `DelegationDistribution` or `DelegationAlignment` or so.

contracts/btc-staking/src/state/delegations.rs
maurolacy: The key does not contain both, just raw.

`rev` is a reverse index of delegations per FP. It can be used to replace the `DELEGATIONS_FPS` store entirely. Same with the direct key, which can be used to replace the current `FP_DELEGATIONS`.

There's already a TODO in the code for that. Will address in a follow-up.

contracts/btc-staking/src/state/points_alignment.rs
maurolacy: Basic test is already there btw. Will work in a multiple delegations test next, to see how the rewards are split amount fps / delegators with different stakes.

contracts/btc-staking/src/staking.rs
maurolacy: Commissions are not implemented yet. I suggest we do it in a follow-up.

contracts/btc-staking/src/state/delegations.rs
maurolacy: Done. We are creating this new struct as it's lighter than the `BtcDelegation`, and relatively independent. We can decide to merge all into one, if we want.

contracts/btc-staking/src/state/delegations.rs
maurolacy: This has been refactored out, as it's not needed in our current impl.

maurolacy:
This is now in a good shape to merge IMO. Please take another look when you find the time.

Follow-ups to this:

- FP commissions.
- Send to Babylon using ICS-020 instead of Bank send (alternative to distribute rewards on Babylon).
- Extra refactors / simplifications:
  - Remove delegation distribution instead of reducing.
  - Use the `delegations()` map and indexes to replace the `DELEGATION_FPS` and `FP_DELEGATIONS` maps.
  - ...

contracts/btc-finality/src/multitest/suite.rs
```diff
@@ -145,6 +151,13 @@ pub struct Suite {
 }
 
 impl Suite {
+    pub fn to_consumer_addr(&self, bbn_addr: &Addr) -> Result<Addr, StakingApiError> {
+        let babylon_prefix = bbn_addr.as_str().split('1').collect::<Vec<_>>()[0];
+        let consumer_prefix = self.babylon.as_str().split('1').collect::<Vec<_>>()[0];
```
SebastianElvis, line 156: Curious why `split('1')`?

contracts/btc-finality/src/contract.rs
```diff
@@ -290,26 +321,34 @@ fn send_rewards_msg(
...
         .collect::<StdResult<Vec<_>>>()?;
-    let msg = babylon_contract::ExecuteMsg::SendRewards {
-        fp_distribution: fp_rewards.clone(),
-    };
-    let wasm_msg = WasmMsg::Execute {
-        contract_addr: cfg.babylon.to_string(),
-        msg: to_json_binary(&msg)?,
-        funds: coins(rewards, cfg.denom.as_str()),
+    let wasm_msg = if recipient == cfg.babylon {
+        let msg = babylon_contract::ExecuteMsg::SendRewards {
```
SebastianElvis, line 331:
```suggestion
        // the reward is sent to the Babylon contract and the Babylon
        // contract will send the reward to Babylon chain for distirbution
        let msg = babylon_contract::ExecuteMsg::SendRewards {
```

contracts/btc-finality/src/contract.rs
```diff
@@ -290,26 +321,34 @@ fn send_rewards_msg(
...
+            fp_distribution: fp_rewards.clone(),
+        };
+        WasmMsg::Execute {
+            contract_addr: recipient.into_string(),
+            msg: to_json_binary(&msg)?,
+            funds: coins(rewards, cfg.denom.as_str()),
+        }
+    } else if recipient == cfg.staking {
+        let msg = btc_staking::msg::ExecuteMsg::DistributeRewards {
```
SebastianElvis, line 340:
```suggestion
        // the reward is sent to the BTC staking contract for distirbution
        let msg = btc_staking::msg::ExecuteMsg::DistributeRewards {
```

contracts/btc-staking/src/staking.rs
```diff
@@ -322,6 +364,154 @@ pub fn handle_slash_fp(
...
 
+pub fn handle_distribute_rewards(
+    mut deps: DepsMut,
+    env: &Env,
+    info: &MessageInfo,
+    rewards: &[RewardInfo],
+) -> Result<Vec<Event>, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
+    // TODO: Check that the sender is the finality contract (AML)
```
SebastianElvis, line 375: Wdyt we implement this in this PR? Seems important as otherwise anyone can trigger reward dist anytime

SebastianElvis (commented):
Still parsing all the math here :+1: Some initial comments, but there does not seem any blocker.

Will take a look at ADR as well

contracts/btc-finality/src/multitest/suite.rs
maurolacy: Because all bech32 addresses have a '1' after the prefix?

This is just a test helper method, didn't check the exact bech32 spec for this.

contracts/btc-finality/src/contract.rs
maurolacy: :+1 This will be changed / deprecated in a follow-up anyway.

contracts/btc-staking/src/staking.rs
maurolacy: Will do.

maurolacy: Done in the latest version.

SebastianElvis (approved):
That's quite some heavylifting, great work!

maurolacy:
Thanks for the support and comments! Will now work in the follow-ups, to put all this into a really good shape.

PR #106, opened 2025-01-31
F/rewards claiming babylon

maurolacy:
Rewards withdrawal to a Babylon address. Depending on the ICS-020 channel information being set or not during contract instantiation.

#102 follow-up / complement. Tracking issue: https://github.com/babylonlabs-io/pm/issues/114.

**TODO**:

- [x] Update relevant [ADR](https://github.com/babylonlabs-io/pm/pull/150).
- [x] Test rewards withdrawal to a Consumer address.
- [x] ~Test rewards withdrawal to a Babylon address~ (to be done as part of e2e tests).

PR #107, opened 2025-02-04
F/rewards claiming babylon user

maurolacy:
#106 follow-up, in which we allow the user to choose the rewards recipient chain (Consumer or Babylon).


SebastianElvis:
Thanks for the effort! According to our previous discussions, it's better to leave the decision of where to claim reward to the consumer rather than individual user. So wdyt we make this PR a draft PR, and mention this alternative design in the ADR?

maurolacy:
Closing. Leaving the branch as reference for future work.

PR #112, opened 2025-02-12
feat: Update consumer BTC light client

gusin13:
1. updates proto definitions
2. new btc_header ibc packet handler to ingest the headers sent from babylon


SebastianElvis (approved):
Great work!

There is conflict though

PR #117, opened 2025-02-19
F/ibc timeout

maurolacy:
Related to https://github.com/babylonlabs-io/pm/issues/165#issuecomment-2668112405.


contracts/babylon/src/ibc.rs
```diff
@@ -110,9 +110,9 @@ pub fn ibc_channel_close(
     // TODO: erase all contract state upon closing the channel
 }
 
-/// Invoked when an IBC packet is received
-/// We decode the contents of the packet and if it matches one of the packets we support
-/// execute the relevant function, otherwise return an error.
+/// Invoked when an IBC packet is received.
+/// We decode the contents of the packet and if it matches one of the packets we support, execute
+/// the relevant function, otherwise return an error.
 /// NOTE: In its current form, this method does not modify state.
```
maurolacy, line 116: @SebastianElvis , is still the case here? What does 'modify state' means btw? Modify state in case of error?

SebastianElvis: Hmm, not the case anymore as this function indeed modifies state

SebastianElvis (approved):

contracts/babylon/src/ibc.rs
maurolacy: Will update the comments, as this is confusing / misleading.

PR #126, opened 2025-03-06
feat: consumer delegation expiry

gusin13:
ADR - https://github.com/babylonlabs-io/pm/pull/263

 -Creates begin block handler in staking contract which handles expiring delegations
-The implementation is bit different from ADR, in ADR i had proposed to process expired delegations in finality contract but that caused some issues so shifted to process delegations in the staking contract itself. The begin blocker handler in staking contract will be triggered by the consumer chain sudo call.

### DIFF
--- a/contracts/btc-staking/src/staking.rs
+++ b/contracts/btc-staking/src/staking.rs
@@ -3,10 +3,11 @@ use bitcoin::consensus::deserialize;
 use bitcoin::hashes::Hash;
 use bitcoin::{Transaction, Txid};
 use cosmwasm_std::{
-    coin, BankMsg, CanonicalAddr, CosmosMsg, DepsMut, Env, Event, IbcMsg, MessageInfo, Response,
-    StdResult, Storage, Uint128, Uint256, Order,
+    BankMsg, CanonicalAddr, CosmosMsg, DepsMut, Env, Event, IbcMsg, MessageInfo,
+    Order, Response, StdResult, Storage, Uint128, Uint256, coin,
 };
 use hex::ToHex;
+use cw_storage_plus::Bound;
 
 use crate::error::ContractError;
 use crate::state::config::{Config, ADMIN, CONFIG, PARAMS};
@@ -28,7 +29,6 @@ use babylon_contract::ibc::packet_timeout;
 use babylon_contract::msg::btc_header::BtcHeaderResponse;
 use babylon_contract::msg::contract::QueryMsg as BabylonQueryMsg;
 use babylon_contract::msg::ibc::TransferInfoResponse;
-use cosmwasm_std::Order::Ascending;
 use cw_utils::{must_pay, nonpayable};
 use std::str::FromStr;
 
@@ -374,11 +374,14 @@ pub fn process_expired_btc_delegations(
     };
 
     // Get all heights that have expired delegations (heights <= current tip height)
-    // We filter for heights that are less than or equal to the current BTC tip height
-    // as these delegations have now expired
+    // Use a bounded range query to efficiently load only the heights we need
     let heights: Vec<u32> = BTC_DELEGATION_EXPIRY_INDEX
-        .keys(deps.storage, None, None, Order::Ascending)
-        .filter(|h| h.as_ref().map_or(false, |&h| h <= tip_height))
+        .keys(
+            deps.storage,
+            None,                               // min bound (start from lowest height)
+            Some(Bound::inclusive(tip_height)), // max bound (up to current tip height)
+            Order::Ascending
+        )
         .collect::<StdResult<Vec<_>>>()?;
 
     // If no expired delegations are found, return early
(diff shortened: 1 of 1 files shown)


contracts/btc-staking/src/contract.rs
```diff
@@ -159,6 +160,28 @@ pub fn execute(
     }
 }
 
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn sudo(
```
SebastianElvis, line 164: I'm wondering whether it's possible to let finality contract calls btc staking contract upon each `BeginBlock`, rather than let Babylon SDK call btc staking contract. This avoids this sudo call, which is imo pretty risky and needs to be minimised

SebastianElvis (commented):
The logic overall LGTM! Only a question on how to trigger BTC staking contract upon BeginBlock

contracts/btc-staking/src/contract.rs
SebastianElvis: Hmm good point, yes the cross-contract call happens only at the end of the message execution. Then I don't think we have the other choice then :+1: 

Wdyt we have some docstring for documenting this?

SebastianElvis (approved):

contracts/btc-staking/src/staking.rs
```diff
@@ -367,6 +352,106 @@ pub fn handle_slash_fp(
...
+            BTC_DELEGATION_EXPIRY_INDEX.remove(deps.storage, btc_height);
+        }
+    }
+
+    Ok(Response::new()
+        .add_attribute("action", "process_expired_delegations")
+        .add_attribute("result", "success"))
+}
+
+fn update_delegation_power(
```
maurolacy, line 415: This function would better be called `discount_delegation_power` or so.

contracts/btc-staking/src/staking.rs
```diff
@@ -230,6 +230,18 @@ pub fn handle_active_delegation(
         ACTIVATED_HEIGHT.save(storage, &(height + 1))?; // Active from the next block onwards
     }
 
+    // Index the delegation by its end height
+    BTC_DELEGATION_EXPIRY_INDEX.update(
```
maurolacy, line 234: This may be inefficient, as at least in principle there could be a good number of delegations by end height.

I think the best way to do this would be to turn the `BTC_DELEGATIONS` `Map` into an `IndexedMap`, and add a multi-index of staking tx hashes per end height.

There are some examples of `MultiIndex` usage in the contracts, but doing this change is involved. Perhaps put a TODO / FIXME in here for now?

maurolacy (commented):
Just some comments. Will review the main logic next / later.

contracts/btc-staking/src/staking.rs
```diff
@@ -367,6 +352,106 @@ pub fn handle_slash_fp(
...
+        }
+    };
+
+    // Get all heights that have expired delegations (heights <= current tip height)
+    // We filter for heights that are less than or equal to the current BTC tip height
+    // as these delegations have now expired
+    let heights: Vec<u32> = BTC_DELEGATION_EXPIRY_INDEX
+        .keys(deps.storage, None, None, Order::Ascending)
+        .filter(|h| h.as_ref().map_or(false, |&h| h <= tip_height))
+        .collect::<StdResult<Vec<_>>>()?;
```
maurolacy, line 382: This is also inefficient / sub-optimal.

There are `range` queries that you can use, instead of `filter`ing over the entire range. Here are some example range queries with bounds:
- https://github.com/babylonlabs-io/babylon-contract/blob/50ba8d7a24cf595a3e9694f66fa0d9f73fda0ed9/contracts/btc-staking/src/queries.rs#L166
- https://github.com/babylonlabs-io/babylon-contract/blob/50ba8d7a24cf595a3e9694f66fa0d9f73fda0ed9/contracts/btc-staking/src/queries.rs#L240

Also, scanning the entire map will become infeasible if there's a large number of entries. It would be better to store the last processed height, and process just the range between the last processed height exclusive, and the current tip height inclusive.

Using an `IndexedMap` you can build the range over the index directly, like is being done here by example: https://github.com/babylonlabs-io/babylon-contract/blob/50ba8d7a24cf595a3e9694f66fa0d9f73fda0ed9/contracts/btc-staking/src/queries.rs#L235-L240.
Let's discuss offline if you want.

contracts/btc-staking/src/staking.rs
```diff
@@ -367,6 +352,106 @@ pub fn handle_slash_fp(
...
+
+                // Only process active delegations
+                if btc_del.is_active() {
+                    // Update delegation power
+                    discount_delegation_power(deps.storage, env.block.height, &staking_tx_hash, &btc_del)?;      
+                }
+            }
+            
+            // Remove the processed height from the index to avoid reprocessing
+            BTC_DELEGATION_EXPIRY_INDEX.remove(deps.storage, btc_height);
```
maurolacy, line 406: 👍🏼 This is good as it'll keep the number of entries at bay.

maurolacy (approved):
Looks good. Let's address the refactor using an indexed map at some point.

PR #128, opened 2025-03-07
F/public randomness ts

### LINKED ISSUE #83: timestamping public randomness support
- [x] ~RFC~ (let's just reference the Babylon's RFC here).
- [x] [ADR](https://github.com/babylonlabs-io/pm/pull/304).
- [x] Implementation.

ADR for timestamping pub rand in Babylon side https://github.com/babylonlabs-io/pm/blob/main/adr/adr-024-timestamping-public-randomness.md

maurolacy:
Closes #83.

### DIFF
--- a/contracts/babylon/src/contract.rs
+++ b/contracts/babylon/src/contract.rs
@@ -216,6 +216,7 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
             &queries::babylon_checkpoint(deps, epoch_number)?,
         )?),
         QueryMsg::CzLastHeader {} => Ok(to_json_binary(&queries::cz_last_header(deps)?)?),
+        QueryMsg::CzLastHeight {} => Ok(to_json_binary(&queries::cz_last_height(deps)?)?),
         QueryMsg::CzHeader { height } => Ok(to_json_binary(&queries::cz_header(deps, height)?)?),
         QueryMsg::TransferInfo {} => Ok(to_json_binary(&queries::transfer_info(deps)?)?),
     }
--- a/contracts/babylon/src/contract.rs
+++ b/contracts/babylon/src/contract.rs
@@ -14,6 +14,7 @@ use crate::msg::contract::{ContractMsg, ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::queries;
 use crate::state::btc_light_client;
 use crate::state::config::{Config, CONFIG};
+use crate::state::cz_header_chain::CZ_HEIGHT_LAST;
 
 pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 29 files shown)


contracts/babylon/src/contract.rs
```diff
@@ -216,6 +224,7 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
             &queries::babylon_checkpoint(deps, epoch_number)?,
         )?),
         QueryMsg::CzLastHeader {} => Ok(to_json_binary(&queries::cz_last_header(deps)?)?),
+        QueryMsg::CzLastHeight {} => Ok(to_json_binary(&queries::cz_last_height(deps)?)?),
```
SebastianElvis, line 227: Why can't we reuse `CzLastHeader`, e.g., let this API return `None` if no header is found?

SebastianElvis (commented):
Great work!

- Shall we add a unit test for this, e.g., if the pub rand corresponding to a finality sig is not finalised, then reject this finality sig?
- Another aspect of timestamped pub rand is that a FP shall have 0 voting power at a height if it does not have timestamped pub rand at this height. This logic is not implemented anywhere, unlike [Babylon](https://github.com/babylonlabs-io/babylon/blob/602f2ed8d0bb27e179a4c25fbd6abdb81de8263d/x/finality/types/power_table.go#L75-L95). Is this planned for the future PRs?

maurolacy:
> Hi @maurolacy I was trying to read code but I lack context. Do we have any ADR somewhere?

Hi'll write ADR next, referencing the original (Babylon) ADR.

maurolacy:
> Shall we add a unit test for this?

Tried to do it, but it's relatively difficult to write a multi-test for it. Perhaps I can write a unit test. Anyway, this is being tested as part of the e2e tests in babylon-sdk.

> Another aspect of timestamped pub rand is that a FP shall have 0 voting power at a height if it does not have timestamped pub rand at this height. This logic is not implemented anywhere, unlike [Babylon](https://github.com/babylonlabs-io/babylon/blob/602f2ed8d0bb27e179a4c25fbd6abdb81de8263d/x/finality/types/power_table.go#L75-L95).

Good point, will take a look and try to do it here. If not I'll create a tracking issue.

contracts/babylon/src/contract.rs
maurolacy: This is just an improvement for performance. It's also more robust, as zero is used to mean "not found" / not set. Finally, this query is simpler to transform into a raw query (again for performance).

maurolacy:
> Is this planned for the future PRs?

Yes, offline (i.e. jailed) and not timestamped FPs will be done as future PRs. Will create the issue about filtering of not timestamped FPs.

SebastianElvis (approved):
:+1: on discounting FPs without timestamped pub rand in a future PR

maurolacy:
Created the relevant issue (https://github.com/babylonlabs-io/babylon-contract/issues/130), and referenced your comment.

PR #133, opened 2025-03-19
fix: handle empty btc headers in timestamp packet

gusin13:
its possible btc timestamp packet has empty/none btc headers - contract shouldn't panic/err


SebastianElvis (approved):
Good catch!

maurolacy (approved):
LGTM.

PR #138, opened 2025-04-15
F/jailing inactive

### LINKED ISSUE #82: jailing inactive finality providers
- [ ] RFC
- [ ] ADR
- [ ] Implementation

maurolacy:
Closes #82.

TODO:

- [x] Unit tests.

### DIFF
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -1,7 +1,7 @@
 use crate::error::ContractError;
 use crate::finality::{
     compute_active_finality_providers, distribute_rewards_fps, handle_finality_signature,
-    handle_public_randomness_commit,
+    handle_jail, handle_public_randomness_commit, handle_unjail,
 };
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state::config::{Config, ADMIN, CONFIG, PARAMS};
@@ -187,6 +187,11 @@ pub fn execute(
             &commitment,
             &signature,
         ),
+        ExecuteMsg::Jail {
+            fp_pubkey_hex,
+            duration,
+        } => handle_jail(deps, &env, &info, &fp_pubkey_hex, duration),
+        ExecuteMsg::Unjail { fp_pubkey_hex } => handle_unjail(deps, &env, &info, &fp_pubkey_hex),
     }
 }
 
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -133,6 +133,9 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
         QueryMsg::Evidence { btc_pk_hex, height } => Ok(to_json_binary(&queries::evidence(
             deps, btc_pk_hex, height,
         )?)?),
+        QueryMsg::JailedFinalityProviders { start_after, limit } => Ok(to_json_binary(
+            &queries::jailed_finality_providers(deps, start_after, limit)?,
+        )?),
     }
 }
 
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -136,6 +136,9 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
         QueryMsg::JailedFinalityProviders { start_after, limit } => Ok(to_json_binary(
             &queries::jailed_finality_providers(deps, start_after, limit)?,
         )?),
+        QueryMsg::ActiveFinalityProviders { height } => Ok(to_json_binary(
+            &queries::active_finality_providers(deps, height)?,
+        )?),
     }
 }
 
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -231,7 +231,7 @@ fn handle_begin_block(deps: &mut DepsMut, env: Env) -> Result<Response<BabylonMs
 
     // Compute active finality provider set
     let max_active_fps = PARAMS.load(deps.storage)?.max_active_finality_providers as usize;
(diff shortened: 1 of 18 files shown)


maurolacy:
> Copilot reviewed 16 out of 16 changed files in this pull request and generated no comments.
> 
> Comments suppressed due to low confidence (1)
> **contracts/btc-finality/src/finality.rs:690**
> 
> * The filtering condition for jailed finality providers defaults to Some(JAIL_FOREVER) on error, which may hide underlying issues by always treating errors as if the FP is jailed. Consider handling the result of may_load explicitly to ensure only truly jailed providers are excluded.
> 
> ```
> JAIL.may_load(deps.storage, &fp.btc_pk_hex).unwrap_or(Some(JAIL_FOREVER)).is_none()
> ```

I know. This is a fallback / default in case there's an error in the underlying storage; which shouldn't happen, or better said, it's catastrophic if it happens (and then this fallback will be utterly unimportant in comparison).

contracts/btc-finality/src/finality.rs
```diff
@@ -588,25 +692,75 @@ pub fn compute_active_finality_providers(
...
+        match last_sign_height {
+            Some(h) if h > env.block.height.saturating_sub(params.missed_blocks_window) => {
+                Ok::<_, ContractError>(())
+            }
+            _ => {
+                // FP is inactive for at least missed_blocks_window, jail! (if not already jailed)
+                JAIL.update(deps.storage, &fp.btc_pk_hex, |jailed| match jailed {
+                    Some(jail_time) => Ok::<_, ContractError>(jail_time),
+                    None => Ok(env.block.time.seconds() + params.jail_duration),
+                })?;
```
SebastianElvis, line 755: Hmm if I understand the design correctly, the jailing condition is as follows: If a FP hasn't submitted `missed_blocks_window` consecutive finality signatures since it becomes active or since its last signature, then the FP will be jailed by the `jail_duration`.

This makes sense to me as a simplified approach, or as a ultimate solution given that we cannot do much heavy-lifting operations in the smart contracts. So maybe we can elaborate this in the ADR

contracts/btc-finality/src/finality.rs
```diff
@@ -291,6 +296,105 @@ pub fn handle_finality_signature(
...
+    // First, ensure the finality provider is jailed
+    let jail_until = JAIL.load(deps.storage, fp_btc_pk_hex)?;
+
+    // Ensure the jail is not forever
+    if jail_until == JAIL_FOREVER {
+        return Err(ContractError::JailedForever {});
+    }
+
+    // Ensure the jail period has passed (except for admin)
+    if !is_admin && env.block.time.seconds() < jail_until {
```
SebastianElvis, line 360: Is this `!is_admin` necessary given that it's checked in L382 anyway?

contracts/btc-finality/src/contract.rs
```diff
@@ -187,6 +193,11 @@ pub fn execute(
             &commitment,
             &signature,
         ),
+        ExecuteMsg::Jail {
+            fp_pubkey_hex,
+            duration,
+        } => handle_jail(deps, &env, &info, &fp_pubkey_hex, duration),
```
SebastianElvis, line 199: Yeah I'd probably avoid this handler given that jailing shall not be bypassed in any way

SebastianElvis (commented):
In general LGTM :+1: will wait for ADR first

packages/apis/src/finality_api.rs
```diff
@@ -49,6 +49,21 @@ pub enum ExecuteMsg {
...
         signature: Binary,
     },
+    /// Jails finality provider. Can be executed only by the admin.
+    Jail {
+        /// FP which should be jailed
+        fp_pubkey_hex: String,
+        /// Duration in seconds for how long finality provider is jailed.
+        /// If it's zero, the FP will be jailed forever.
+        duration: u64,
+    },
```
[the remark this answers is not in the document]
maurolacy: For administrative (whatever they are) reasons? I don't know, can remove if you think it's useless / overkill.

contracts/btc-finality/src/test_utils.rs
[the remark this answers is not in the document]
maurolacy: Sure. This is adapted / copied from the staking contract, btw.

contracts/btc-finality/src/finality.rs
maurolacy: This is good enough as a jailing condition for being offline, IMO. The Babylon offline detection is more complex, as it considers a certain percentage of blocks over a given period, or blocks window.

Saying, for example 10% of missed blocks over a 1000 blocks window, is (almost) the same as saying, 1 missed block over a 10 blocks window. The more complex condition can be reduced to a simpler / unitary one, while keeping the same ratio.
I don't think we miss much with this design, and its simplicity is considerably greater.

contracts/btc-finality/src/finality.rs
maurolacy: It is, if we want admins to be able of unjail even before the jailing period is over (which makes sense for an admin).

contracts/btc-finality/src/contract.rs
maurolacy: Do you mean, removing the ability of an admin to jail? We can remove it, yes.

packages/apis/src/finality_api.rs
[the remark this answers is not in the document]
maurolacy: This comes from some tgrade contracts / examples, where the admin has this power

https://github.com/confio/poe-contracts/blob/b7a8dbafd89cd70401dced518366f520b7089ff6/contracts/tgrade-valset/src/contract.rs#L194-L196

contracts/btc-finality/src/contract.rs
SebastianElvis: yep i don't see it's needed (for testing or for production)

SebastianElvis (approved):
:rocket:

PR #140, opened 2025-04-24
F/filter non timestamped

### LINKED ISSUE #130: filter out non-timestamped FPs from active FP list
After #83, let's add a filter over FPs that don't have timestamped public randomness, so that they are excluded from the active FP set.

This is similar to filtering out jailed (#82) FPs.

Some references / context:

- [Babylon FP filtering](https://github.com/babylonlabs-io/babylon/blob/602f2ed8d0bb27e179a4c25fbd6abdb81de8263d/x/finality/types/power_table.go#L75-L95).
- @SebastianElvis comment (https://github.com/babylonlabs-io/babylon-contract/pull/128#pullrequestreview-2688500240): _It's worth nothing to mention that, voting power table is better maintained in the finality contract rather than BTC staking contract. After we have timestamped pub rand and jailing, the voting power of a FP does not only depend on the total active BTC stake, but also timestamped pub rand and jailing. Both info are only stored in the finality contract. If we keep voting power table in BTC staking contract then there will be circular dependency between BTC staking contract and finality contract._

So, implementing this properly and efficiently may be tricky.

maurolacy:
Closes #130.

### DIFF
--- a/contracts/btc-finality/src/state/public_randomness.rs
+++ b/contracts/btc-finality/src/state/public_randomness.rs
@@ -78,20 +78,18 @@ pub fn has_timestamped_pub_rand_commit_for_height(
     fp_btc_pk_hex: &str,
     height: u64,
     last_finalized_height: Option<u64>,
-) -> Result<bool, ContractError> {
-    let pr_commit = get_pub_rand_commit_for_height(deps.storage, fp_btc_pk_hex, height)?;
+) -> bool {
+    let pr_commit = match get_pub_rand_commit_for_height(deps.storage, fp_btc_pk_hex, height) {
+        Ok(pr_commit) => pr_commit,
+        Err(_) => return false,
+    };
 
     // Ensure the finality provider's corresponding randomness commitment is already finalised by
     // BTC timestamping
-    let finalized_height = last_finalized_height.unwrap_or(get_last_finalized_height(deps)?);
-    if finalized_height == 0 {
-        return Ok(false);
-    }
-    if finalized_height < pr_commit.height {
-        return Ok(false);
-    }
+    let finalized_height =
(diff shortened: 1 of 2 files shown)


contracts/btc-finality/src/state/public_randomness.rs
```diff
@@ -70,6 +70,30 @@ pub fn get_timestamped_pub_rand_commit_for_height(
...
+// `has_timestamped_pub_rand_commit_for_height` checks that the FP has timestamped public
+// randomness commitment for the given height and (optionally given) last finalized height.
+// If the last finalized height is not given, it will be queried from the babylon contract.
+pub fn has_timestamped_pub_rand_commit_for_height(
+    deps: &Deps,
+    fp_btc_pk_hex: &str,
+    height: u64,
+    last_finalized_height: Option<u64>,
+) -> Result<bool, ContractError> {
+    let pr_commit = get_pub_rand_commit_for_height(deps.storage, fp_btc_pk_hex, height)?;
```
SebastianElvis, line 82: Why not we

- return a `bool` only without handling the error, and
- return `false` if `get_pub_rand_commit_for_height` returns an error?

SebastianElvis (approved):
LGTM :+1:

contracts/btc-finality/src/state/public_randomness.rs
gusin13: yep agree with Runchao's point, I think it'll be similar to how done in Babylon then. 

https://github.com/babylonlabs-io/babylon/blob/d204f9535d6d2f426a5e5956d4c5987be865419d/x/finality/keeper/public_randomness.go#L39-L62

gusin13 (approved):
nice work, lgtm! ✅ 

sorry for delay in review

PR #163, opened 2025-06-29
fix: avoid output filename collision warning

liuchengxu:
Currently, you'll see the following warning when running `cargo build`:

```
warning: output filename collision.
The bin target `schema` in package `btc-light-client v0.14.0 (/Users/xuliucheng/src/github.com/babylonlabs-io/cosmos-bsn-contracts/contracts/btc-light-client)` has the same output filename as the bin target `schema` in package `babylon-contract v0.14.0 (/Users/xuliucheng/src/github.com/babylonlabs-io/cosmos-bsn-contracts/contracts/babylon)`.
Colliding filename is: /Users/xuliucheng/src/github.com/babylonlabs-io/cosmos-bsn-contracts/target/debug/schema
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The bin target `schema` in package `btc-light-client v0.14.0 (/Users/xuliucheng/src/github.com/babylonlabs-io/cosmos-bsn-contracts/contracts/btc-light-client)` has the same output filename as the bin target `schema` in package `babylon-contract v0.14.0 (/Users/xuliucheng/src/github.com/babylonlabs-io/cosmos-bsn-contracts/contracts/babylon)`.
Colliding filename is: /Users/xuliucheng/src/github.com/babylonlabs-io/cosmos-bsn-contracts/target/debug/schema.dSYM
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

This PR fixes it by renaming the schema binary to be contract-specific. It also fixes the schema generator in `deploy.yml.`

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -102,11 +102,7 @@ jobs:
         asset_name: checksums.txt
         asset_content_type: text/plain
     - name: Build and run schema generator
-      run: |-
-        (cd ./contracts/babylon && cargo run --bin schema)
-        (cd ./contracts/btc-staking && cargo run --bin btc-staking-schema)
-        (cd ./contracts/btc-finality && cargo run --bin btc-finality-schema)
-        (cd ./contracts/btc-light-client && cargo run --bin btc-light-client-schema)
+      run: bash schema.sh
     - name: Consolidate schemas
       run: |-
         mkdir -p ./schemas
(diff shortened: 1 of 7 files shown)


SebastianElvis (approved):
Good catch!

Issue #103, opened 2025-01-16
Flatten instantiation config / params 


maurolacy:
We're using optional config / params fields for contract instantiation (https://github.com/babylonlabs-io/babylon-contract/blob/9fdc9a15de1f082bc0a0686fe4be6bf320a03a53/contracts/btc-staking/src/msg.rs#L16) and providing default values through `derivative` (in case the entire config / params field is not set) (https://github.com/babylonlabs-io/babylon-contract/blob/9fdc9a15de1f082bc0a0686fe4be6bf320a03a53/contracts/btc-staking/src/state/config.rs#L23-L52).

This may be good for simplicity, but it's bad for ergonomics. If we want to override a single default value, we need to provide the entire `Config` / `Params` struct fields (if not, `serde` will complain during deserialisation).

So, it's better to make those config / params fields explicit, flattening them in the `InstantiationMsg` directly. That way they are visible, settings are clear and well documented, there's  no surprising / default behaviour, and less eventual head scratching.

We can still use default values through `derivative` or helpers, but let's restrict those to tests, and test helpers.

This will be breaking contract instantiation in a number of places (e2e tests, integration scripts, etc.).
But, that's better than implicitly setting undesired / unknown default values.

Regarding `Config` vs. `Params` state / storage, we could simplify that as well putting everything under the same struct. But I think the separation has some merit: config values are those immutable values gathered during instantiation, whereas we can use params for mutable / settable values (through governance, administrator accounts, etc.)

SebastianElvis:
Good point, we should do this

PR #164, opened 2025-06-29
refactor: improve babylon contract instantiation

liuchengxu:
This PR refactors the `InstantiateMsg` in the Babylon contract by grouping initialization parameters for the BTC light client, BTC staking, and BTC finality contracts into dedicated nested structs:

- `BtcLightClientInitMsg`
- `BtcStakingInitMsg`
- `BtcFinalityInitMsg`

This clarifies which fields belong to which module, resulting in a more structured and readable configuration. The validation of the staking message is also simplified.

This change updates the contract’s instantiation interface, which affects CLI usage. Given the clarity gained, I believe this is acceptable, but let me know if it breaks any expected assumptions.

Early feedback is appreciated before I fix the tests.


contracts/babylon/src/msg/contract.rs
```diff
@@ -26,36 +51,18 @@ pub struct InstantiateMsg {
...
-    pub btc_finality_msg: Option<Binary>,
-    /// If set, this will be the Wasm migration / upgrade admin of the BTC staking contract and the
-    /// BTC finality contract
+    /// Initialization info for the BTC light client contract.
+    pub btc_light_client_init_msg: Option<BtcLightClientInitMsg>,
+    /// Initialization info for the BTC staking contract.
+    pub btc_staking_init_msg: Option<BtcStakingInitMsg>,
+    /// Initialization info for the BTC finality contract.
+    pub btc_finality_init_msg: Option<BtcFinalityInitMsg>,
+    /// Wasm migration / upgrade admin of the BTC staking contract and the BTC finality contract.
```
SebastianElvis, line 64: I recommend we avoid bundling code ID and instantiation message and keep these struct as is. They are separate things. Also, `InstantiateMsg` will have to be submitted as a JSON string via CLI, and CosmWasm provides poor support for nested JSON.

contracts/babylon/src/msg/contract.rs
```diff
@@ -18,6 +18,31 @@ pub trait ContractMsg {
...
+    pub btc_light_client_msg: Option<Binary>,
+}
+
+#[cw_serde]
+pub struct BtcStakingInitMsg {
+    pub btc_staking_code_id: u64,
+    /// Name of the consumer
+    pub consumer_name: String,
+    /// Description of the consumer
+    pub consumer_description: String,
```
SebastianElvis, line 34: The concepts `consumer_name` and `consumer_description` exist at system level, not just in the BTC staking contract. So it belongs to Babylon contract (which is the "router" contract)

liuchengxu: That's fair, then this PR does not make sense.

contracts/babylon/src/msg/contract.rs
liuchengxu: I saw that the nested structure has been used in https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/49ea2a81de1b430afeb1951e4925ef9b4834529e/contracts/btc-finality/src/msg.rs#L15, so this is supposedly fine, I guess. Anyway, your other comment already renders this PR invalid, closing.

maurolacy:
Likely related issue you might want to work / address: https://github.com/babylonlabs-io/cosmos-bsn-contracts/issues/103.

Issue #166, opened 2025-06-30
Introduce code style guidelines


liuchengxu:
Currently, the code style across this repo isn’t well-defined, which can lead to inconsistencies and confusion during development. It would be great to introduce a basic style guide, something like [bitvm-bridge’s CONTRIBUTING.md](https://github.com/babylonlabs-io/bitvm-bridge/blob/main/CONTRIBUTING.md#code-style), to improve readability and streamline collaboration going forward.

Here are a few initial suggestions (non-exhaustive):

- Import consistency: Right now, import statements vary quite a bit—some files group everything together, others split them up in different ways without a super clear pattern. I'd recommend we adopt a consistent style across the board. Personally, I prefer a single group of imports for simplicity (no strong opinion though, if you have any other preferences, please just propose one and I'll follow).

- Redundant comments: Comments with redundant prefixes are unnecessary and can be removed to reduce noise. For example, https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/49ea2a81de1b430afeb1951e4925ef9b4834529e/contracts/babylon/src/msg/consumer_header.rs#L14-L15, the `consumer_id` in the comment restates the obvious and doesn’t add value.

- `unwrap()` usage: Since this is contract code and runs in a blockchain context, we should avoid bare `unwrap()`, you can instead use expect() with clear justifications for why failure is impossible or acceptable in that context.

Happy to contribute a first pass at the style guide if you agree. @SebastianElvis @maurolacy

maurolacy:
My logic for organising imports, that I inherited from CosmWasm repos, is as follows: From most general to most specific imports, top to bottom. With an empty line to separate blocks.

This means, at the top go crates that have general usage (`serde`, `hex`, etc.). Then (typically) CosmWasm crates. Then local packages; then local cross-contract imports; then local imports (`state`, `msg`, etc.). 

I agree, this logic is currently not implemented over all of the imports.

Issue #165, opened 2025-06-29
Rename consumer to BSN in codebase


liuchengxu:
Once the renaming to BSN is finished on the contracts side, we may also do the same renaming on the babylon-sdk side.

PR #167, opened 2025-06-30
refactor: introduce CONTRIBUTING.md

liuchengxu:
This PR introduces the initial CONTRIBUTING.md for the code style guide. The changes in this PR primarily address the comment and `unwrap()` part mentioned in #166 .

### DIFF
--- a/contracts/babylon/src/bindings.rs
+++ b/contracts/babylon/src/bindings.rs
@@ -1,4 +1,5 @@
-//!`bindings` include functions that invoke custom messages to the Consumer chain.
+//! This module includes functions that invoke custom messages to the Consumer chain.
+
 use cosmwasm_std::StdError;
 
 use babylon_bindings::BabylonMsg;
--- a/contracts/babylon/src/contract.rs
+++ b/contracts/babylon/src/contract.rs
@@ -144,12 +144,17 @@ pub fn reply(
     _env: Env,
     reply: Reply,
 ) -> Result<Response<BabylonMsg>, ContractError> {
+    let response = || {
+        reply
+            .result
+            .into_result()
+            .expect("TODO: why it's okay to not handle error here")
+    };
+
     match reply.id {
-        REPLY_ID_INSTANTIATE_LIGHT_CLIENT => {
-            reply_init_callback_light_client(deps, reply.result.unwrap())
-        }
-        REPLY_ID_INSTANTIATE_STAKING => reply_init_callback_staking(deps, reply.result.unwrap()),
-        REPLY_ID_INSTANTIATE_FINALITY => reply_init_callback_finality(deps, reply.result.unwrap()),
+        REPLY_ID_INSTANTIATE_LIGHT_CLIENT => reply_init_callback_light_client(deps, response()),
+        REPLY_ID_INSTANTIATE_STAKING => reply_init_callback_staking(deps, response()),
+        REPLY_ID_INSTANTIATE_FINALITY => reply_init_callback_finality(deps, response()),
         _ => Err(ContractError::InvalidReplyId(reply.id)),
     }
 }
--- a/contracts/babylon/src/msg/consumer_header.rs
+++ b/contracts/babylon/src/msg/consumer_header.rs
@@ -3,7 +3,7 @@ use cosmwasm_schema::cw_serde;
 use cosmwasm_std::Timestamp;
 use hex::ToHex;
 
-/// ConsumerHeaderResponse is the metadata of a Consumer header.
+/// Metadata of a Consumer header.
 ///
 /// This struct is for use in RPC requests and responses. It has convenience helpers to convert
 /// from the internal representation (`IndexedHeader`).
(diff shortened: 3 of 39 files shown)


contracts/babylon/src/contract.rs
```diff
@@ -144,12 +144,17 @@ pub fn reply(
     _env: Env,
     reply: Reply,
 ) -> Result<Response<BabylonMsg>, ContractError> {
+    let response = || {
+        reply
+            .result
+            .into_result()
+            .expect("TODO: why it's okay to not handle error here")
```
liuchengxu, line 151: This is for you :P @SebastianElvis @maurolacy

liuchengxu: If I understand correctly, what you mean is that when the `reply()` function is invoked, the `reply.result` must be `Ok(SubMsgResponse)`, otherwise `reply()` function won't be called at all. Then I'm curious why not define a new `Reply` that includes the response directly:

```Rust
pub struct NewReply {
    pub id: u64,
    pub payload: Binary,
    pub gas_used: u64,
    pub sub_msg_response: SubMsgResponse,
}
```

The comment on the existing `Reply` seems to be fallible:

>The result object returned to reply. We always get the ID from the submessage back and then must handle success and error cases ourselves.

Ref https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.Reply.html

Just wanted to know what happens here as your explanation seems to conflict with the comment on `Reply`.

SebastianElvis: > Then I'm curious why not define a new Reply that includes the response directly:

I remember there is some function called `reply_on_success` but we did not use that.

> We always get the ID from the submessage back and then must handle success and error cases ourselves.

Hmm right, I think this indicates that we have to handle the error cases. For some reason this was not the case before. I think we will have to handle the error here then

contracts/babylon/src/msg/consumer_header.rs
```diff
@@ -3,41 +3,41 @@ use cosmwasm_schema::cw_serde;
...
+/// Metadata of a Consumer header.
 ///
 /// This struct is for use in RPC requests and responses. It has convenience helpers to convert
 /// from the internal representation (`IndexedHeader`).
 ///
 /// Adapted from `IndexedHeader`.
 #[cw_serde]
 pub struct ConsumerHeaderResponse {
-    /// consumer_id is the unique ID of the consumer
+    /// Unique ID of the consumer
```
maurolacy, line 14: Go tell Golang devs about this. :-P

contracts/btc-finality/src/contract.rs
```diff
@@ -71,9 +71,12 @@ fn get_blocks_per_year(deps: &mut DepsMut) -> Result<u64, ContractError> {
...
         )?;
         // Deserialize protobuf
-        let res_decoded = anybuf::Bufany::deserialize(&res).unwrap();
+        let res_decoded =
+            anybuf::Bufany::deserialize(&res).expect("Static value must not fail to deserialize");
         // See https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/proto/cosmos/mint/v1beta1/query.proto#L35-L36
-        let res_params = res_decoded.message(1).unwrap();
+        let res_params = res_decoded
+            .message(1)
+            .expect("Static value must contain a message at index 1");
```
maurolacy, line 79: Better. Another (better) option is to add a `ProtobufDecode` from `BufanyError` contract error, and return it here through `?`.

maurolacy (approved):
LGTM.

Personally I agree with the doc change.
Just mentioning that in many cases this comes from Go conventions / comments, which follow a different convention.

contracts/babylon/src/bindings.rs
```diff
@@ -1,4 +1,5 @@
-//!`bindings` include functions that invoke custom messages to the Consumer chain.
+//! This module includes functions that invoke custom messages to the Consumer chain.
```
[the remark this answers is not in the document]
liuchengxu: Let's take care of this in #165

contracts/btc-staking/src/state/delegations.rs
```diff
@@ -41,7 +41,7 @@ pub struct Delegations<'a> {
 
 impl Delegations<'_> {
     fn deserialize_pk(pk: &[u8]) -> (Vec<u8>, String) {
-        <(Vec<u8>, String)>::from_slice(pk).unwrap() // mustn't fail
+        <(Vec<u8>, String)>::from_slice(pk).expect("TODO: why must not fail")
```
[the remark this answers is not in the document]
liuchengxu: That's fair, then we could just say something like `deserialization must not fail otherwise the store is broken`, letting the reader know what's happening here.

contracts/babylon/src/contract.rs
liuchengxu: Hey Mauro, can I leave these two TODOs for you and merge this PR sooner?

PR #171, opened 2025-07-01
refactor: remove `ics23-commitment` fork

liuchengxu:
The `ics23-commitment` fork is no longer needed, we can use `ibc-core-commitment-types` now.


SebastianElvis (approved):
Great!

PR #174, opened 2025-07-02
F/publish crates io

maurolacy:
Publish package crates to crates.io.

So that tehy can be used from other repos, i.e. rollup-bsn-contracts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,6 +283,7 @@ version = "0.14.0"
 dependencies = [
  "babylon-bitcoin",
  "babylon-proto",
+ "babylon-test-utils",
  "bitcoin",
  "digest",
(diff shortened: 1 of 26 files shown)


liuchengxu (approved):
I'm wondering if it might make more sense to just use a Git dependency on the rollup contract side for now, since we expect cosmos-bsn-contracts to go through significant and rapid changes in the coming weeks before the testnet.

maurolacy:
> I'm wondering if it might make more sense to just use a Git dependency on the rollup contract side for now, since we expect cosmos-bsn-contracts to go through significant and rapid changes in the coming weeks before the testnet.

For now, yes. This is to address versioning issues in the near future.

PR #176, opened 2025-07-02
feat: initialize btc-light-client on babylon contract instantiation

liuchengxu:
This PR refactors the BTC light client's instantiation process. Instead of being initialized in an empty state, it now requires an initial list of headers to begin its operation.

**Main changes:**

   - The btc-light-client contract's `InstantiateMsg` is updated to accept a list of headers, a base work value, and a starting height.
   - The Babylon contract is updated to provide the initial headers when instantiating the light client.
   - The representation of the `work` attribute has been standardized to use raw bytes (Binary or Vec<u8>) across contracts, removing the previous hex-encoded string format.
   - Introduced a `new_test` constructor for the Babylon contract's `InstantiateMsg` to simplify test setup.

Reviewing commit by commit might make more sense.

**TODO:**

The complete validation of the initial headers is left as a TODO for the follow-up. I will revisit it when I work on the missing consensus verification.

Part of https://github.com/babylonlabs-io/babylon/issues/1240#issuecomment-2982569008


packages/test-utils/src/lib.rs
```diff
@@ -341,3 +349,35 @@ pub fn get_public_randomness_commitment() -> (String, PubRandCommit, Vec<u8>) {
...
+/// Returns the initial BTC headers encoded in protobuf and hex-encoded form.
+///
+/// This format is suitable for embedding into messages such as `InstantiateMsg`.
+pub fn initial_headers_in_hex() -> String {
+    let mut out = Vec::new();
+    for h in initial_headers() {
+        h.encode_length_delimited(&mut out).unwrap();
+    }
+    hex::encode(out)
+}
```
SebastianElvis, line 383: Hmm why not let `InstantiateMsg` take an array of hex strings?

contracts/btc-light-client/src/msg/contract.rs
```diff
@@ -13,20 +13,37 @@ pub struct InstantiateMsg {
     pub network: babylon_bitcoin::chain_params::Network,
     pub btc_confirmation_depth: u32,
     pub checkpoint_finalization_timeout: u32,
+    /// A sequence of initial BTC headers used to bootstrap the light client.
+    ///
+    /// Must include at least `btc_confirmation_depth` headers following the base header.
+    pub headers: Vec<BtcHeader>,
```
SebastianElvis, line 19: Do you think we should make it

```suggestion
    /// Must include at least `btc_confirmation_depth` headers following the base header.
    pub headers: Vec<String>,
```

such that one can instantiate BTC light client contract through CLI?

contracts/btc-light-client/src/msg/contract.rs
```diff
@@ -13,20 +13,37 @@ pub struct InstantiateMsg {
...
+    /// Total accumulated work of the first (base) BTC header, encoded as big-endian bytes.
+    pub first_work: Binary,
```
SebastianElvis, line 21: Same question here, maybe hex is more friendly to CLI?

contracts/babylon/src/msg/contract.rs
```diff
@@ -32,6 +32,8 @@ pub struct InstantiateMsg {
     pub notify_cosmos_zone: bool,
     /// If set, this will instantiate a BTC light client contract
     pub btc_light_client_code_id: Option<u64>,
+    /// Encoded `Vec<BtcHeaderInfo>` in hex.
+    pub btc_light_client_initial_headers: String,
```
SebastianElvis, line 36: Do we need first work and first height here?

contracts/btc-light-client/src/msg/contract.rs
SebastianElvis: Btw, an alternative here is to take a single `BtcHeaderInfo` for instantiation. The header is picked carefully by the BSN so we don't need to have `k` deep ones. Also if the BSN picks a bad one Babylon won't be synchronising BTC headers with this BSN.

contracts/babylon/src/msg/contract.rs
liuchengxu: Nope, as you can see from the previous code, the first work and height are derived from the initial base header. This is consistent with the old logic.

contracts/btc-light-client/benches/main.rs
```diff
@@ -45,10 +45,14 @@ pub fn get_main_msg_test_headers() -> Vec<BtcHeader> {
 #[track_caller]
 pub fn setup_instance() -> Instance<MockApi, MockStorage, MockQuerier> {
     let mut deps = mock_instance_with_gas_limit(WASM, 10_000_000_000_000);
+    let initial_headers = initial_headers();
     let msg = InstantiateMsg {
         network: babylon_bitcoin::chain_params::Network::Regtest,
         btc_confirmation_depth: 10,
         checkpoint_finalization_timeout: 2,
```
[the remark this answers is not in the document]
liuchengxu: No logical changes in this file and I didn't write this test initially, you or Mauro wrote it :P

contracts/btc-light-client/src/msg/contract.rs
liuchengxu: I thought about this. So do we still need to instantiate the contract from CLI? I thought we don't need the instantiation from CLI, hence I chose the simplicity here. Do you have any scenarios that we may instantiate the LC contract directly from CLI instead of going through the babylon contract?

>Also if the BSN picks a bad one Babylon won't be synchronising BTC headers with this BSN.

I'm not familar with this part. More input is appreciated.

packages/test-utils/src/lib.rs
liuchengxu: No particular reason, I just used `String` initially. I can update it if using an array of hex strings is preferred.

contracts/btc-light-client/src/msg/contract.rs
SebastianElvis: > Do you have any scenarios that we may instantiate the LC contract directly from CLI instead of going through the babylon contract?

Not in our context, but at some point people might take this BTC light client contract for their own use cases. The contract is rather self-contained so I thought we can make it generic

> Also if the BSN picks a bad one Babylon won't be synchronising BTC headers with this BSN.
> I'm not familar with this part. More input is appreciated.

So basically after this PR, Babylon side will also be updated such that upon establishing IBC channel, the contract can send its initial BTC header to Babylon, and Babylon will start forwarding BTC headers since that header. If the BSN chooses an initial header that is invalid, Babylon won't be able to determine the list of BTC headers for forwarding. It's BSN's responsibility to pick a BTC header that is valid and is recent.

That's why we might be able to have a single initial header for simplicity

liuchengxu: Switched to the single initial header approach, please have another look. A bunch of comments are outdated now and I have resolved them.

contracts/babylon/src/msg/contract.rs
```diff
@@ -32,6 +32,8 @@ pub struct InstantiateMsg {
     pub notify_cosmos_zone: bool,
     /// If set, this will instantiate a BTC light client contract
     pub btc_light_client_code_id: Option<u64>,
+    /// JSON encoded `InitialHeader` in hex.
+    pub btc_light_client_initial_header: String,
```
SebastianElvis, line 36: `btc_light_client_msg` (the instantiation message for BTC light client) already contains the initial header so probably we don't need this field

SebastianElvis (approved):

packages/test-utils/src/lib.rs
```diff
@@ -339,3 +347,27 @@ pub fn get_public_randomness_commitment() -> (String, PubRandCommit, Vec<u8>) {
         pub_rand_commitment_msg.sig.to_vec(),
     )
 }
+
+/// Returns the initial BTC header for the babylon contract instantiation.
+pub fn initial_header() -> btc_light_client::msg::contract::InitialHeader {
```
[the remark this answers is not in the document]
liuchengxu: I've thought about this — I haven't found a better home for `InitialHeader` as ideally I'd prefer to avoid adding more code into the `packages` and I want to slim the `packages` in the future if possible. Since `test-utils` is only used in the test environment, I think it's acceptable for now. Let's keep it in mind and see if we can do better in future.

contracts/babylon/src/msg/contract.rs
liuchengxu: It's necessary when `btc_light_client_msg` is `None`.

liuchengxu: >It's necessary when btc_light_client_msg is None

I was trying to keep the original logic that attempted to construct an `InstantiateMsg` when `btc_light_client_msg` was `None`, but now I realize this is unnecessary after discussing with Gai. If `btc_light_client_msg` is None, we may error out immediately rather than trying to handle it implicitly.

- Before: If `btc_light_client_msg` is None, attempt to construct an `InstantiateMsg`.
- After: If `btc_light_client_msg` is None, return an error.

PR #180, opened 2025-07-06
feat(btc-light-client): rework header verification

### LINKED ISSUE #137: btclc: tighten the BTC light client verification rule

liuchengxu:
This PR reworks the header verification, incorporating the difficulty adjustment check. I have verified this PR against the Babylon Genesis implementation, and they should now match closely.

Tagging @KonradStaniec since you implemented this in Babylon Genesis — would be great to get your 👀 on it :P

Close #137

### DIFF
--- a/contracts/btc-light-client/src/bitcoin.rs
+++ b/contracts/btc-light-client/src/bitcoin.rs
@@ -3,7 +3,7 @@
 use crate::error::ContractError;
 use crate::state::get_header;
 use babylon_bitcoin::{deserialize, BlockHeader, Work};
-use babylon_bitcoin::{Params, Uint256 as U256};
+use babylon_bitcoin::{Params, Uint256};
 use babylon_proto::babylon::btclightclient::v1::BtcHeaderInfo;
 use bitcoin::Target;
 use cosmwasm_std::Storage;
@@ -102,11 +102,9 @@ fn check_block_header_sanity(
     // if the chain does not allow reduced difficulty after 10min, ensure
     // the new header's target is within the [0.25, 4] range
     if !chain_params.allow_min_difficulty_blocks {
-        let retarget_adjustment_factor_u256 =
-            cosmwasm_std::Uint256::from(RETARGET_ADJUSTMENT_FACTOR);
-        let old_target =
-            cosmwasm_std::Uint256::from_be_bytes(prev_block_header.target().to_be_bytes());
-        let cur_target = cosmwasm_std::Uint256::from_be_bytes(header.target().to_be_bytes());
+        let retarget_adjustment_factor_u256 = Uint256::from(RETARGET_ADJUSTMENT_FACTOR);
+        let old_target = Uint256::from_be_bytes(prev_block_header.target().to_be_bytes());
+        let cur_target = Uint256::from_be_bytes(header.target().to_be_bytes());
         let max_cur_target = old_target * retarget_adjustment_factor_u256;
         let min_cur_target = old_target / retarget_adjustment_factor_u256;
         if cur_target > max_cur_target || cur_target < min_cur_target {
@@ -200,7 +198,7 @@ fn get_next_work_required(
         let last_block_time = last_block.time;
 
         Ok(calculate_next_work_required(
-            U256::from_be_bytes(last_block.target().to_be_bytes()),
+            Uint256::from_be_bytes(last_block.target().to_be_bytes()),
             first_block_time.into(),
             last_block_time.into(),
             params,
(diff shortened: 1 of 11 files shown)


contracts/btc-light-client/src/bitcoin.rs
```diff
@@ -1,54 +1,306 @@
...
+    if !chain_params.allow_min_difficulty_blocks {
+        let retarget_adjustment_factor_u256 = Uint256::from(RETARGET_ADJUSTMENT_FACTOR);
+        let old_target = Uint256::from_be_bytes(prev_block_header.target().to_be_bytes());
+        let cur_target = Uint256::from_be_bytes(header.target().to_be_bytes());
+        let max_cur_target = old_target * retarget_adjustment_factor_u256;
+        let min_cur_target = old_target / retarget_adjustment_factor_u256;
+        if cur_target > max_cur_target || cur_target < min_cur_target {
+            return Err(HeaderError::BadDifficulty);
+        }
+    }
```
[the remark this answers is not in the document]
SebastianElvis: This is legacy difficulty check and shall be subsumed by the one implemented in this PR

contracts/btc-light-client/src/bitcoin.rs
```diff
@@ -1,54 +1,277 @@
...
+        storage,
+        chain_params,
+        prev_block_height,
+        prev_block_header,
+        header,
+    )?;
+
+    // Perform proof-of-work check as specified in:
+    // https://pkg.go.dev/github.com/btcsuite/btcd@v0.24.2/blockchain#CheckBlockHeaderSanity
+    check_proof_of_work(chain_params, header)?;
```
gitferry, line 134: It seems the sanity check includes both pow check and timestamp check. Why do we skip timestamp check here?

liuchengxu: See this comment https://github.com/babylonlabs-io/babylon/blob/3e474546ab80b2696a66c10506f84bb17ce53bca/types/btcutils.go#L17-L21

contracts/btc-light-client/src/bitcoin.rs
```diff
@@ -1,54 +1,277 @@
...
+            last_block_time.into(),
+            params,
+        ))
+    } else {
+        Ok(last_block.target())
+    }
+}
+
+// <https://github.com/bitcoin/bitcoin/blob/89b910711c004c21b7d67baa888073742f7f94f0/src/pow.cpp#L49-L72>
+fn calculate_next_work_required(
```
KonradStaniec, line 240: maybe we can re-use functions from rust bitcoin - https://github.com/rust-bitcoin/rust-bitcoin/blob/f7274a57c36d0b8d8cc528426e269c27d496bd5f/bitcoin/src/pow.rs#L374 ?  They should be compatible with `cpp` and `golang` impls.

contracts/btc-light-client/src/bitcoin.rs
```diff
@@ -1,54 +1,277 @@
...
+
+/// > Ensure the difficulty specified in the block header matches
+/// > the calculated difficulty based on the previous block and
+/// > difficulty retarget rules.
+///
+/// Note: While the naming mirrors the btcd implementation, this function only performs
+/// the difficulty adjustment check (not the full header context validation in btcd).
+///
+/// https://pkg.go.dev/github.com/btcsuite/btcd@v0.24.2/blockchain#CheckBlockHeaderContext
+fn check_block_header_context(
```
KonradStaniec, line 172: the context check has also this median time check:
```go
		// Ensure the timestamp for the block header is after the
		// median time of the last several blocks (medianTimeBlocks).
		medianTime := CalcPastMedianTime(prevNode)
		if !header.Timestamp.After(medianTime) {
			str := "block timestamp of %v is not after expected %v"
			str = fmt.Sprintf(str, header.Timestamp, medianTime)
			return ruleError(ErrTimeTooOld, str)
		}
	}
```
I wonder whether missing it may lead to some discrepancies or accepting headers that should not be accepted 🤔

KonradStaniec (commented):
In general it looks good 👍 

My main concern is that maybe we can re-use some functions from https://github.com/rust-bitcoin/rust-bitcoin (similary like Babylon re-uses btcd) ?

contracts/btc-light-client/src/bitcoin.rs
liuchengxu: Good catch! This slipped through because I got a bit mixed up between the future timestamp above and the median time check. We should definitely add this check.

contracts/btc-light-client/src/bitcoin.rs
liuchengxu: If I remember correctly, it was a bit inconvenient to reuse the APIs from rust-bitcoin at the time. But that might not be the case anymore. I can take another look and see if things work smoothly now.

contracts/btc-light-client/src/bitcoin.rs
liuchengxu: Synced with @gitferry offline, let's merge this PR on his approval for the upcoming devnet. I'll add the median time check in a follow-up PR shortly.

PR #181, opened 2025-07-06
feat(btc-light-client): check if the initial header on the difficulty boundary

### LINKED ISSUE #136: initialisation of BTC light client in consumer (key point: 2016-block epoch in BTC)

liuchengxu:
Close #136

### DIFF
--- a/contracts/btc-light-client/src/msg/contract.rs
+++ b/contracts/btc-light-client/src/msg/contract.rs
@@ -59,8 +59,16 @@ impl InstantiateMsg {
             return Err(ContractError::ZeroCheckpointFinalizationTimeout);
         }
 
-        // TODO: validate headers, first work and first height? For example, the base header
-        // should be on the difficulty boundary.
+        if !crate::bitcoin::is_difficulty_change_boundary(
+            self.initial_header.height,
+            &self.network.chain_params(),
+        ) {
+            return Err(ContractError::NotOnDifficultyBoundary(
+                self.initial_header.height,
+            ));
+        }
+
+        // TODO: the height should be larger than a recent block?
 
         Ok(())
     }
--- a/packages/test-utils/src/lib.rs
+++ b/packages/test-utils/src/lib.rs
@@ -350,9 +350,9 @@ pub fn get_public_randomness_commitment() -> (String, PubRandCommit, Vec<u8>) {
 
 /// Returns the initial BTC header for the babylon contract instantiation.
 pub fn initial_header() -> btc_light_client::msg::contract::InitialHeader {
-    // https://www.blockchain.com/explorer/blocks/btc/354816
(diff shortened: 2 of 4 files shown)


contracts/btc-light-client/src/msg/contract.rs
```diff
@@ -59,8 +59,16 @@ impl InstantiateMsg {
...
+        if !crate::bitcoin::is_difficulty_change_boundary(
+            self.initial_header.height,
+            &self.network.chain_params(),
+        ) {
+            return Err(ContractError::NotOnDifficultyBoundary(
+                self.initial_header.height,
+            ));
+        }
+
+        // TODO: the height should be larger than a recent block?
```
SebastianElvis, line 71: Hmm it's BSN's responsibility to pick a recent height. If the BSN picks a height that is very old then the block sync will have some issues. So maybe this TODO is not needed

packages/test-utils/src/lib.rs
```diff
@@ -350,9 +350,12 @@ pub fn get_public_randomness_commitment() -> (String, PubRandCommit, Vec<u8>) {
...
 pub fn initial_header() -> btc_light_client::msg::contract::InitialHeader {
-    // https://www.blockchain.com/explorer/blocks/btc/354816
-    let header = "020000003f99814a36d2a2043b1d4bf61a410f71828eca1decbf56000000000000000000b3762ed278ac44bb953e24262cfeb952d0abe6d3b7f8b74fd24e009b96b6cb965d674655dd1317186436e79d";
-    let height = 354816;
+    // Initial base header on Babylon Genesis mainnet, https://www.blockchain.com/explorer/blocks/btc/854784.
+    // TODO: This hardcodes a mainnet header in `initial_header()`, which may be incorrect in a
+    // different network context, and we do often use different networks (e.g., testnet or regtest) in the test environment.
+    // It's fine for now, but we should make this function network-aware to avoid subtle bugs down the line.
+    let header = "0000c020f382af1f6d228721b49f3da2f5b831587803b16597b301000000000000000000e4f76aae64d8316d195a92424871b74168b58d1c3c6988548e0e9890b15fc2fc3c00aa66be1a0317082e4bc7";
+    let height = 854784;
```
SebastianElvis, line 358: At some point we could have a simple Rust CLI to return the suggested instantiation message template, with a recent header in the diff adjustment boundary. But this is not high pri at this moment

SebastianElvis (approved):
Maybe not for this PR, but a unit test could be good to have

contracts/btc-light-client/src/msg/contract.rs
liuchengxu: That’s fair. I was just thinking it might be nice to catch some obvious misconfigurations in the BSN setup, hence the TODO with a question mark :( We can remove it for simplicity.

PR #190, opened 2025-07-10
feat(btc-light-client): check median time

liuchengxu:
This PR follows up on #180 by adding the median time check to the Bitcoin light client. cc @KonradStaniec


contracts/btc-light-client/src/bitcoin.rs
```diff
@@ -193,9 +214,70 @@ fn check_block_header_context(
...
+    // BIP 113
+    let height = prev_block_height + 1;
+
+    let csv_height = match chain_params.network {
+        Network::Bitcoin => 419328, // 000000000000000004a1b34462cb8aeebd5799177f7a29cf28f2d1961716b5b5
+        Network::Testnet => 770112, // 00000000025e930139bac5c6c31a403776da130831ab85be56578f3fa75369bb
+        Network::Signet => 1,       // Always active unless overridden
+        Network::Regtest => 1,      // Always active unless overridden
+        _ => unreachable!("Unsupported network"),
+    };
```
vitsalis, line 226: These height checks are not part of `CheckBlockHeaderContext`, why do we add them here?

vitsalis (commented):
One comment. Could we also have some test cases for the new functionality?

contracts/btc-light-client/src/bitcoin.rs
liuchengxu: Thanks for pointing this out! I got confused by BIP 113, which defines the transaction locktime constraint using the median time past. The CSV height check is only required when validating that constraint, which we don't need in the light client, so it's unnecessary in this context.

liuchengxu (commented):
>Could we also have some test cases for the new functionality?

I had actually added some tests locally in #180, but I wasn’t entirely happy with them, so I didn’t include them in the PR. Ideally, I wanted to port existing test cases from Bitcoin Core or other implementations, but that didn’t go as smoothly as I hoped. In any case, I’ve now added some tests with the help of an AI fork, let’s see how they hold up :P

contracts/btc-light-client/src/bitcoin.rs
```diff
@@ -193,9 +215,60 @@ fn check_block_header_context(
...
+            break;
+        }
+
+        let header: BlockHeader = match get_header_by_hash(storage, block_hash.as_ref()) {
+            Ok(raw_header) => deserialize(raw_header.header.as_ref())?,
+            Err(_) => pending_headers
+                .get(&block_hash)
+                .map(|(header, _height)| header)
+                .cloned()
+                .ok_or(HeaderError::MissingHeader(block_hash))?,
```
KonradStaniec, line 255: Does it mean if header is missing it will return error? do we have some check in contract that there is always at least 11 headers in store (for example after initialisation of contract) ?

KonradStaniec (commented):
there is another check which is missing from contract which is in golang version. There is this check that header has proper version based on some upgrades heights:
```go
	params := c.ChainParams()
	if header.Version < 2 && blockHeight >= params.BIP0034Height ||
		header.Version < 3 && blockHeight >= params.BIP0066Height ||
		header.Version < 4 && blockHeight >= params.BIP0065Height {

		str := "new blocks with version %d are no longer valid"
		str = fmt.Sprintf(str, header.Version)
		return ruleError(ErrBlockVersionTooOld, str)
	}
```

Given that those versions are super old, maybe it is just enough that header has version `4` 🤔

contracts/btc-light-client/src/bitcoin.rs
liuchengxu: You're right, we should use as many headers as available, rather than erroring when fewer than 11 are present. I've rewritten the `calculate_median_time_past()` function accordingly. This change does require a conceptual shift in how we interact with storage.

Notably, the semantic of `get_header_by_hash()` has been updated to return `Result<Option<T>>` instead of `Result<T>`. This is because `None` can be a valid outcome in certain situations. Currently, many  of our storage APIs return an error when a value is missing, which might not always be semantically correct. Let me know if you have any thoughts on `get_header_by_hash()` and `expect_header_by_hash()`. cc @SebastianElvis

KonradStaniec: >Notably, the semantic of get_header_by_hash() has been updated to return Result<Option<T>> instead of Result<T>. This is because None can be a valid outcome in certain situations. Currently, many of our storage APIs return an error when a value is missing, which might not always be semantically correct. Let me know if you have any thoughts on get_header_by_hash() and expect_header_by_hash()

👍 it makes sense to distinguish error and lack of element on type level

PR #201, opened 2025-07-12
chore: update datagen methods to use Babylon v3

vitsalis:


SebastianElvis:
Have we tried to generate new test data and re-run all test cases?

I'm pretty sure with the new test data, some tests will be broken.

vitsalis:
@SebastianElvis ah thought the data was generated automatically. Could someone pick this up?

PR #207, opened 2025-07-13
refactor: simplify btc staking contract

liuchengxu:
This PR is a pure refactor of the btc-staking contract in preparation for #7. The error handling are simplified a lot and now it's super clear what's the difference between the full and lite version of the validation after moving the full validation to a separate module. While there’s still room for further cleanup and modularization, I’m capping the scope here and starting to focus on implementing the remaining full-validation features. Further refinements may follow once the feature set stabilizes.


gitferry (approved):
Nice refactoring!

PR #220, opened 2025-07-18
chore: remove `btc_light_client_initial_header`

liuchengxu:
@SebastianElvis was right that `btc_light_client_initial_header` is unnecessary as it's already contained in the `btc_light_client_msg`. See https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/176#discussion_r2215338173 for more info.

### DIFF
--- a/contracts/babylon/src/contract.rs
+++ b/contracts/babylon/src/contract.rs
@@ -7,9 +7,6 @@ use cw_utils::ParseReplyError;
 
 use babylon_apis::{btc_staking_api, finality_api};
 use babylon_bindings::BabylonMsg;
-use btc_light_client::msg::contract::{
-    InitialHeader, InstantiateMsg as BtcLightClientInstantiateMsg,
-};
 
 use crate::error::ContractError;
 use crate::ibc::{ibc_packet, IBC_CHANNEL, IBC_TRANSFER};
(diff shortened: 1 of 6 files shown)


maurolacy:
There seem to be a lot of confusion going on with this BTC light client initialisation stuff:

- Initial header is being passed twice. This supposedly fixes that, but initial header is still there in the Babylon contract `Config` entry.
- Using a tuple to store info is counter-intuitive and error prone. And, it breaks a lot of stuff along the line. It also makes this field incompatible with the other ones for btc-staking and btc-finality, which are simple addresses.
- Sending these headers to Babylon reverses the flow, and introduces a lot of coupling.
- Requiring a different initial header for BSN's btc light clients break a lot of working e2e tests.

I would say: Why on Earth do we want BSNs to have an initial header at all? Can't they just stick to the same BTC Babylon uses for initial header? What are the advantages? Do they warrant the significant amount of extra complexity and coupling?

Let's discuss offline, in any case.

liuchengxu:
I would say as a self-contained BTC light client implementation, it's critical to have the initial header defined within the contract. Letting the Babylon Genesis handle the initialization seems simpler for the short-term, but loses the reusablity in another context. I don't understand why this introduces that much complexity, this design is prevalent in the existing BTC light client implementations.

maurolacy:
> I don't understand why this introduces that much complexity,

It breaks a lot of things, and has limited to no utility in the current use case. Making it optional will at least allow for backwards compatibility.

SebastianElvis:
As per offline discussion, we will need to implement the initial BTC header logic. The reason is security

> The problem in the current "sending last w headers" design is security. Suppose Babylon has adversarial majority, the Babylon chain can just send IBC packets with any well-formatted BTC headers and associated BTC timestamps and BTC staking txs. BSN side will blindly accept the BTC headers and all subsequent BTC timestamps / staking txs. However, these BTC headers could be a fork mined through little voting power.
>
> To solve this, BSN has to pick a set of parameters upon instantiation, including a recent BTC header on Bitcoin main chain. This prevents adversarial Babylon chain from inserting fork BTC headers to BSN chain.

SebastianElvis (approved):

PR #223, opened 2025-07-21
fix(btc-light-client): store the base header data

### LINKED ISSUE #222: BTC base header is not correctly stored in the contract instantiation
Currently, we store the base header in BTC_BASE_HEADER and update the tip, but we fail to insert the base header into BTC_HEADERS and BTC_HEIGHTS. This omission causes submission of the next header to fail, as it cannot find its parent — the base header — in storage. While it exists in BTC_BASE_HEADER, the rest of the header-related state is missing, breaking the expected storage invariants.

This issue stems from a mismatch introduced during the migration from the old initialization logic to the current design.

https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/6df345aaa6cc4e2aad833f722d4e0793ebf2b3d5/contracts/btc-light-client/src/state/btc_light_client.rs#L265

https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/6df345aaa6cc4e2aad833f722d4e0793ebf2b3d5/contracts/btc-light-client/src/contract.rs#L46-L47

https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/6df345aaa6cc4e2aad833f722d4e0793ebf2b3d5/contracts/btc-light-client/src/state/btc_light_client.rs#L115-L120

To resolve this:

- Replace `BTC_HEADER_BASE` (which stores the full base header) with `BTC_BASE_HEADER_HEIGHT`, storing only the base height.
- Extract a new `insert_header()` helper from `insert_headers()`, handling insertion into all relevant maps: `BTC_HEADERS`, `BTC_HEIGHTS`, and others.
- Use this `insert_header()` for both base header insertion and normal header processing, ensuring uniform treatment and consistent storage layout.

This unification reduces code duplication and eliminates the risk of partial/inconsistent state when inserting headers in the future.

liuchengxu:
Gai discovered this bug while working on updating the contracts e2e tests in babylon-sdk, see #222 for more context.

Close #222

One thing I noticed: Currently, we store the entire header for the BTC tip, which might be unnecessary as we already have the tip header stored in the `BTC_HEADERS` storage. We may only store the metadata (header_hash, header_height) for the BTC tip.

### DIFF
--- a/contracts/btc-light-client/src/contract.rs
+++ b/contracts/btc-light-client/src/contract.rs
@@ -51,11 +51,8 @@ pub fn instantiate(
     // Set contract version
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
 
-    let mut buf = Vec::with_capacity(base_header.encoded_len());
-    base_header.encode(&mut buf)?;
-
     Ok(Response::new()
-        .set_data(Binary::from(buf))
+        .set_data(Binary::from(base_header.encode_to_vec()))
         .add_attribute("action", "instantiate"))
 }
 
--- a/contracts/btc-light-client/src/lib.rs
+++ b/contracts/btc-light-client/src/lib.rs
@@ -16,6 +16,8 @@ pub mod error;
 pub mod msg;
 pub mod queries;
 pub mod state;
+#[cfg(test)]
+mod tests;
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
--- a/contracts/btc-light-client/src/state/btc_light_client.rs
+++ b/contracts/btc-light-client/src/state/btc_light_client.rs
@@ -15,8 +15,8 @@ use super::CONFIG;
 
 pub const BTC_TIP_KEY: &str = "btc_lc_tip";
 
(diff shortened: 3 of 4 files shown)


maurolacy:
Having the e2e tests in `babylon-sdk` is fine.

What is important is, updating and making those e2e tests pass after every breaking change. That is, updating `babylon-sdk` every time a breaking feature is introduced in the contracts. The same with the `babylon` repo, and the reference to it in `babylon-sdk` (i.e. the babylon docker image there).

That way we make sure everything runs smoothly, and don't have to face a mountain of breaking changes during release.

liuchengxu:
>What is important is, updating babylon-sdk and making e2e tests pass after every breaking change. That is, updating babylon-sdk every time a breaking feature is introduced in the contracts.

This setup worked well in the past, but the situation is a bit different now that Gai is focused on `babylon-sdk` and I'm owning the contracts repo. I can write Go, but Gai's apparently much faster at it. Ideally, we should each double down on what we do best to maximum efficiency and avoid context-switching overhead. In my opinion, having the e2e in the contracts repo can potentially make us move faster with fewer bugs.

PR #225, opened 2025-07-22
chore: include the btc_staking-full-validation.wasm in the release

liuchengxu:
The full validation of btc staking contract is compiled in the pipeline but not include in the release artifacts.

A bit more thinking: Since now we are clear that we'll directly deploy the full validation version, perhaps it makes sense to remove the `full-validation` feature and make it default.


gitferry:
I'd say we keep it until production ready in case we change the plan in the middle

PR #228, opened 2025-07-23
feat: Optional btc lc initial header

maurolacy:
Makes the `btc-light-client` initial BTC header optional, and dependent on the `full-validation` feature.


maurolacy:
> The old initialization logic had been removed but not yet recovered though.

How come? Maybe I missed something. Can you point to sections in the code base?

> updating the Rust and CI can be a separate PR which will make reviewer's life easier.

Good point. My local was failing and I decided to fix this in passing, sorry.

maurolacy:
Just removed the CI / clippy stuff from this PR.

Will publish in a follow-up PR.

maurolacy:
Good catch! Will add that to recover the btc-lc auto-init / catch-up logic.

maurolacy:
This has suffered significant refactoring (which is wasn't aware of).

In particular, the introduction of the (optional) `first_height` and `first_work` headers.

@SebastianElvis, can you explain the rationale behind those fields, and if we should restore them? What is the problem with the original version, in which if the headers were not initialised, the first batch of vectors served as initialisation (i.e. the `init_from_user` approach).

Which behaviour should we restore for lc auto init of headers?
Thanks.

gitferry (approved):
Lgtm!

PR #230, opened 2025-07-23
update: Rust version in CI

maurolacy:
#228 dependent / addition.

Also fixes / silences clippy warnings.


liuchengxu (approved):
I suggest we also add `rust-toolchain.toml` so that we can always have a consistent clippy result with CI.

Issue #241, opened 2025-07-24
EPIC: misalignment in BTC light client


liuchengxu:
During the implementation of the difficulty adjustment feature for the BTC light client contract, my validation primarily focused on header verification by cross-referencing the Go implementation. We now need to conduct a full-scope cross-check.

PR #242, opened 2025-07-24
refactor(btc-light-client): simplify `btc-light-client` contract structure

liuchengxu:
The previous structure of the btc-light-client contract was fragmented, with logic unncessarily spread across multiple files and modules. This made it difficult to navigate the codebase and understand the relationships between different components. This pure refactoring addresses these issues by creating a more logical and organized structure, which will make the contract easier to maintain and develop in the future IMHO.

**Main changes:**

   - _Consolidated State Management_: The various state-related modules (`state/{btc_light_client, config, mod}.rs`) have been consolidated into a single `state.rs` file. This simplifies state management and makes the code easier to follow.
   - _Unified Query Module_: The query logic has been unified into a single `queries.rs` file, removing the nested module structure.
   - _Centralized Execution Logic_: The core execution logic for handling BTC headers has been moved into the main contract.rs file. This includes functions like `handle_btc_headers_from_babylon`, `init_btc_headers`, and `extend_btc_headers`.
   - _Other Minor Refactorings_: The contract instantiation logic has been simplified, code cleanup, etc.
   
If this PR looks good to you, I propose we apply the same to the other contracts when feasible. Please share any feedback or concerns, and I'm happy to discuss further.

This refactoring will make it easier for me to work on #241.


contracts/btc-light-client/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+    #[error("The bytes cannot be decoded")]
+    Decode(#[from] prost::DecodeError),
+    #[error(transparent)]
+    CosmwasmStd(#[from] StdError),
+    #[error("The BTC height {height} is not found in the storage")]
+    HeightNotFound { height: u32 },
+    #[error("The BTC header with hash {hash} is not found in the storage")]
+    HeaderNotFound { hash: String },
+}
+
```
SebastianElvis, line 32: Do these need to be under `error.rs` or here? Same question for other errors

SebastianElvis (approved):
Nice!

contracts/btc-light-client/src/state.rs
liuchengxu: Not necessarily, I'd say. I don't think we have to define all the Error types in `error.rs`. It depends on the context. For a large crate-level Error type, we could either:

- Split it into smaller, more focused error types, or
- Encapsulate them hierarchically within relevant modules.

The right choice comes down to maintainability, whichever keeps the codebase clean and easier to reason about. For smaller crate Error, a single error type may suffice, but for larger ones, modularizing errors often helps.

Issue #7, opened 2024-07-31
EPIC: Add full validations


maurolacy:
**[<img src="https://avatars.githubusercontent.com/u/11656534?s=17&v=4" width="17" height="17"> @maurolacy](https://www.github.com/maurolacy)** cloned issue [babylonchain/babylon-contract#84](https://www.github.com/babylonchain/babylon-contract/issues/84) on 2024-05-03: 

> After basic validations (#82), let's complete the validation logic after the original Go code.
> 
> 
> 
> Related to #82. Also, likely requires #76.

SebastianElvis:
After #46 and #56, most of the reality checks are done (feature flag for lite/full version, correctness of BTC staking library, etc..).

The following includes the missing verification rules. Those will be picked up later.

## Finality provider registration request

- [ ] Verify PoP

## BTC delegation request

- [ ] Ensure staking tx is included the BTC light client
- [ ] Ensure staking tx is k-deep
- [ ] Ensure staking tx timelock has more than w BTC blocks left
- [ ] Verify PoP
- [ ] Verify covenant signatures over slashing tx
- [ ] Ensure unbonding time is max(min_unbonding_time, w)
- [ ] Ensure unbonding tx is spending the given staking tx
- [ ] Ensure consistency bweteen unbonding tx and unbonding slashing tx
- [ ] Verify staker's signature over slashing path of unbonding tx
- [ ] Verify covenant signatures over unbonding slashing tx
- [ ] Ensure unbonding tx fee is not too low

## BTC undelegation request

- [ ] Ensure BTC delegation is active
- [ ] Verify staker's signature over the unbonding tx

## Slashed BTC delegation

- [ ] Ensure BTC delegation is active
- [ ] Verify the evidence of the equivocating FP

maurolacy:
There are a number of TODOs in the apis `validate` module. Listing here for visibility:

- https://github.com/babylonlabs-io/babylon-contract/blob/50ba8d7a24cf595a3e9694f66fa0d9f73fda0ed9/packages/apis/src/validate.rs#L29-L31.
- https://github.com/babylonlabs-io/babylon-contract/blob/50ba8d7a24cf595a3e9694f66fa0d9f73fda0ed9/packages/apis/src/validate.rs#L33-L37.
[...]

PR #244, opened 2025-07-24
chore: remove PoP from validation

SebastianElvis:
Part of https://github.com/babylonlabs-io/cosmos-bsn-contracts/issues/7

This PR removes PoP from validation because it's not necessary

### DIFF
--- a/contracts/btc-staking/src/validation/full.rs
+++ b/contracts/btc-staking/src/validation/full.rs
@@ -30,7 +30,9 @@ pub enum FullValidationError {
 
 /// Verifies the new finality provider data.
 /// TODO: Implement validation logic for the new finality provider.
-pub fn verify_new_fp(_new_fp: &NewFinalityProvider) -> Result<(), ContractError> {
+pub fn verify_new_fp(new_fp: &NewFinalityProvider) -> Result<(), ContractError> {
+    verifying_key_from_hex(&new_fp.btc_pk_hex)?;
+
     Ok(())
 }
 


liuchengxu (approved):

contracts/btc-staking/src/validation/full.rs
```diff
@@ -37,49 +29,8 @@ pub enum FullValidationError {
...
-            return Ok(());
-        }
-        BTCSigType::ECDSA => {
-            // TODO?: implement ECDSA verification (#7.0)
-            return Ok(());
-        }
-    }
-
+/// TODO: Implement validation logic for the new finality provider.
+pub fn verify_new_fp(_new_fp: &NewFinalityProvider) -> Result<(), ContractError> {
```
maurolacy, line 33: Why not leave the FP key and staker address validation?

What are the requirements for validating new FPs on the BSN side?

contracts/btc-staking/src/validation/full.rs
```diff
@@ -37,49 +29,8 @@ pub enum FullValidationError {
...
-
-    // FIXME: parameterise `bbn` prefix
-    let fp_address = to_canonical_addr(&new_fp.addr, "bbn")?;
-
-    let fp_pop = new_fp
-        .pop
-        .as_ref()
-        .ok_or(FullValidationError::MissingProofOfPossession)?;
-
-    verify_pop(&fp_pk, fp_address, fp_pop)?;
```
maurolacy, line 51: Is this not necessary in both lite and full versions?

PR #248, opened 2025-07-25
[DNM] BTC light client lite version

SebastianElvis:


gitferry:
This pr still makes sense, right? For lite version, we don't need to verify headers

PR #249, opened 2025-07-25
fix(btc-light-client): rename `is_difficulty_boundary()` to `is_retarget_block()` and fix the misalignment

liuchengxu:
This fixes the misalignment in the difficulty block adjustment check. Can't recall why `height >= difficulty_adjustment_interval` was added previously, anyway, now it aligns with the Go implementation, with the reference attached in the comment. Not 100% if it's conceptually right that the block 0 is also a retarget block, but it's aligned at least :(


SebastianElvis (approved):
CI failed though

PR #255, opened 2025-07-28
Feat: Transfer info not optional

### LINKED ISSUE #226: ICS transfer channel id should not be optional during instantiation
Currently, ICS transfer channel id is optional during instantiation due to old rewards protocol design. Given that new design forces rewards dist to happen on Babylon Genesis, the ICS transfer channel id must be set during instantiation

maurolacy:
Closes #226.


maurolacy:
We should. Low priority though.

PR #257, opened 2025-07-29
btclc: trusted version of the BTC light client

### LINKED ISSUE #259: Remove base_header completely from btc lc contract instantiation
base_header should not be set in light of trusting Babylon Genesis with lite verification on contracts

SebastianElvis:
Resolves #259 

This PR provides the trusted version of the BTC light client. It 

- removes the work check,
- removes the base header in instantiation message
- enforces sender of the BTC headers is either contract admin or the Babylon contract

### DIFF
--- a/contracts/btc-light-client/schema/btc-light-client.json
+++ b/contracts/btc-light-client/schema/btc-light-client.json
@@ -12,15 +12,10 @@
       "network"
     ],
     "properties": {
-      "base_header": {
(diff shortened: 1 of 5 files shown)


liuchengxu (approved):
I guess the removal of header verification will be another PR?

SebastianElvis:
> I guess the removal of header verification will be another PR?

The remaining verification will be kept as per offline discussion. In particular, we are still keeping the checks of consecutive headers and fork choices. Those are kept as basic sanity check. If those checks failed then this means something is deeply wrong in Babylon side.

Issue #263, opened 2025-07-29
Missalignment in finality votes verification


gitferry:
The following should be filled in https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/1eb5086b78ce3fa6aa25d32d40e5548c5f3030b1/contracts/btc-finality/src/finality.rs#L150
 
1. [x] missing stateless validation in the add finality message https://github.com/babylonlabs-io/babylon/blob/80d89b10add5d914f2a7353b725b803b17fb7cc5/x/finality/keeper/msg_server.go#L68
2. [x] reject the finality sig early if the block is already finalized https://github.com/babylonlabs-io/babylon/blob/80d89b10add5d914f2a7353b725b803b17fb7cc5/x/finality/keeper/msg_server.go#L82 (#288)
3. [x] upon duplicate vote, error should be returned https://github.com/babylonlabs-io/babylon/blob/80d89b10add5d914f2a7353b725b803b17fb7cc5/x/finality/keeper/msg_server.go#L132
4. ~~missing fundable feature~~ https://github.com/babylonlabs-io/babylon/blob/80d89b10add5d914f2a7353b725b803b17fb7cc5/x/finality/keeper/msg_server.go#L240 (we decided to not support refunding)

gitferry:
@SebastianElvis do you think (4) is a missing feature?

SebastianElvis:
> do you think (4) is a missing feature?

I think we have decided to NOT support refunding

gitferry:
Note that (3) is api-breaking and fp should adapt it

liuchengxu:
@SebastianElvis Regarding the point 2 `reject the finality sig early if the block is already finalized`, two places in Go impl are using `CheckpointingKeeper`, i.e., 	`epochNum := ms.CheckpointingKeeper.GetEpochByHeight(ctx, block.Height)
	lastFinalizedEpoch := ms.GetLastFinalizedEpoch(ctx)`,

- I can find the `get_last_finalized_epoch()` which is in `babylon-contract`. I need to do a contract query from btc-finality?
- I can't find the alternative of `get_epoch_by_height()`. I guess this is missing right now and I should add it?

SebastianElvis:
[...]

PR #265, opened 2025-07-30
finality: handle public randomness misalignment

### LINKED ISSUE #262: Misalignments in handling public randomness
1. miss stateless checks on public randomness message fields https://github.com/babylonlabs-io/babylon/blob/7bcb1085c67b6a6447d78facbab1a82a829a8f67/x/finality/types/msg.go#L161
2. miss check of randomness length https://github.com/babylonlabs-io/babylon/blob/80d89b10add5d914f2a7353b725b803b17fb7cc5/x/finality/keeper/msg_server.go#L267

liuchengxu:
This PR aims to resolve the misalignment listed in #262.

One important thing in this PR is the introduction of the `PublicRandomnessCommitMsg` type which mirrors the original Go implementation. I strongly recommend we do the same to the others, i.e., we only unpack the fields in the tx interface layer, once the msg crosses the boundary, we should immediately convert them into the types like `PublicRandomnessCommitMsg`, making it easier to reason about and align with the original Go impl and audit.

TODO:

- [ ] Port the Go tests.

Close #262


contracts/btc-finality/src/finality.rs
```diff
@@ -24,19 +24,126 @@ use k256::sha2::{Digest, Sha256};
...
+pub struct PublicRandomnessCommitMsg {
+    pub fp_btc_pk_hex: String,
+    pub start_height: u64,
+    pub num_pub_rand: u64,
+    pub commitment: Vec<u8>,
+    pub sig: Vec<u8>,
+}
+
+impl PublicRandomnessCommitMsg {
+    // https://github.com/babylonlabs-io/babylon/blob/49972e2d3e35caf0a685c37e1f745c47b75bfc69/x/finality/types/msg.go#L161
```
gitferry, line 57: maybe better to put the ref on the top of PublicRandomnessCommitMsg as the ref covers the entire impl

gitferry (approved):
Nice work!

contracts/btc-finality/src/finality.rs
liuchengxu: You meant the entire impl of `PublicRandomnessCommitMsg` structure? That's not true, this permalink is specifically for the function `ValidateBasic`.

liuchengxu (commented):
Merging this and starting to work on #263. Will revisit it to port the Go tests.

PR #274, opened 2025-07-31
feat: Rewards to Babylon

### LINKED ISSUE #227: Stage the rewards from BSN and forward them to Babylon Genesis periodically
We plan to implement the BSN reward transfer for the distribution on Babylon Genesis within the contract. Here is the flow on the contract side:

- Reward collection per block.
    - The BSN sends rewards (e.g., a portion of collected fees) to the contract on every block.
    - These rewards are accumulated in a dedicated pool within the contract.
- Periodic Forwarding to Babylon Genesis
    - Instead of transferring rewards immediately, they are batched and forwarded periodically (e.g., once per epoch).
    - The forwarding interval will be a configurable parameter.

cc @gitferry @SebastianElvis @maurolacy

maurolacy:
Closes #227.


contracts/babylon/src/contract.rs
```diff
@@ -310,15 +311,80 @@ pub fn execute(
...
+            }
+
+            // Get the ICS20 transfer channel ID
+            let channel_id = IBC_TRANSFER.load(deps.storage)?;
+
+            // Create memo with distribution information as JSON
+            let memo = serde_json::json!({
+                "fp_distribution": fp_distribution
+            })
+            .to_string();
```
gitferry, line 348: The memo constructed should match the structure defined in the Babylon Genesis

gitferry (commented):
In general, I think we are heading the right direction. One quick comment:

contracts/babylon/src/contract.rs
maurolacy: Yes, taking a look at that now.

contracts/babylon/src/contract.rs
```diff
@@ -309,15 +311,80 @@ pub fn execute(
...
+            let channel_id = IBC_TRANSFER_CHANNEL.load(deps.storage)?;
+
+            // Create memo with distribution information as JSON
+            let memo = serde_json::json!({
+                "fp_distribution": fp_distribution
+            })
+            .to_string();
+
+            // Define destination address
+            let rcpt_addr = to_bech32_addr("bbn", &to_module_canonical_addr("zoneconcierge"))?;
```
SebastianElvis, line 351: The destination is BTC staking module not zone concierge module https://github.com/babylonlabs-io/babylon/blob/3e0b24b9063362b1387eec6647116fb722d8d0cd/x/btcstaking/types/ibc_callbacks.go. Also, would be great if we can parameterise it

contracts/btc-finality/src/contract.rs
```diff
@@ -228,10 +233,14 @@ fn handle_update_staking(
 }
 
 fn handle_begin_block(deps: &mut DepsMut, env: Env) -> Result<Response, ContractError> {
+    // Distribute rewards to finality providers
+    distribute_rewards_fps(deps, &env)?;
```
SebastianElvis, line 237:
```suggestion
    // Distribute rewards to finality providers
    record_fp_rewards(deps, &env)?;
```

The function is only doing accounting but not actual reward distribution

contracts/babylon/src/ibc.rs
```diff
@@ -159,7 +159,7 @@ pub(crate) mod ibc_packet {
 
     use super::*;
 
-    fn get_ibc_packet_timeout(env: &Env, deps: &Deps) -> StdResult<IbcTimeout> {
+    pub fn get_ibc_packet_timeout(env: &Env, deps: &Deps) -> StdResult<IbcTimeout> {
         let cfg = CONFIG.load(deps.storage)?;
         let timeout = env.block.time.plus_days(cfg.ibc_packet_timeout_days);
         Ok(IbcTimeout::with_timestamp(timeout))
```
SebastianElvis, line 165: Maybe this function can be moved out from this sub crate

contracts/babylon/src/contract.rs
```diff
@@ -445,4 +542,195 @@ mod tests {
...
+            notify_cosmos_zone: false,
+            btc_light_client: None,
+            btc_staking: None,
+            btc_finality: Some(finality_addr.clone()),
+            consumer_name: Some("TestConsumer".to_string()),
+            consumer_description: Some("Test Consumer Description".to_string()),
+            denom: "stake".to_string(),
+            ibc_packet_timeout_days: 1,
+            destination_module: "btcstaking".to_string(),
+        };
```
liuchengxu, line 565: This may be extracted as a small helper function or method of `Config`.

contracts/babylon/src/contract.rs
gitferry: Yes, you are right 👍

contracts/babylon/src/contract.rs
```diff
@@ -309,15 +311,109 @@ pub fn execute(
             // TODO: Add events (#124)
             Ok(res)
         }
+        ExecuteMsg::DistributeRewards { fp_distribution } => {
```
liuchengxu, line 314: How about `SendRewardsDistribution` or something similar to better clarify that it does not perform the distribution but forwards the distribution info to Babylon Genesis?

contracts/babylon/src/contract.rs
```diff
@@ -309,15 +311,109 @@ pub fn execute(
...
+                    Ok(FpRatio {
+                        btc_pk: Binary::new(btc_pk_bytes),
+                        ratio: reward_decimal,
+                    })
+                })
+                .collect::<Result<Vec<_>, ContractError>>()?;
+
+            // Create memo with the proper Babylon structure using typed structs
+            let callback_memo = CallbackMemo {
+                action: "add_bsn_rewards".to_string(),
```
liuchengxu, line 366: Not an IBC expert, I'm wondering if it's a convention to name the action `add_bsn_rewards` matching the info in `dest_callback` or if it must be named as such; otherwise, the callback won't work.

contracts/babylon/src/contract.rs
```diff
@@ -309,15 +311,109 @@ pub fn execute(
...
+            response = response.add_message(transfer_msg);
+
+            Ok(response)
+        }
     }
 }
 
 #[cfg(test)]
 mod tests {
+
```
liuchengxu, line 409:
```suggestion
```

liuchengxu (approved):
A few minor comments, otherwise looks good on my side.

gitferry (commented):
Another non-blocking comment is that, do we really need the `epoch_length` as parameter? What value should we put there? And, maybe we should rename it to to something like `reward_interval` to not confuse people with the epoch concept in Babylon Genesis

contracts/babylon/src/contract.rs
maurolacy: This is part of the (custom) memo field defintion. It has to be set to that value (which could be a constant) for the message to work / be processed.

contracts/btc-finality/src/contract.rs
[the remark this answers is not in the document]
gitferry: I don't think it makes sense. If no blocks are finalized, we should not distribute rewards? Even with our current fee collector rewards, we still should be able to do this

Issue #268, opened 2025-07-30
Misalignment in btc-staking contracts queries


gitferry:
1. [finality provider](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/9fff11cf8c31523fbe3df75491af9f81ad90064f/contracts/btc-staking/src/contract.rs#L59): remove description and commission from the response and db as we don't have edit events to keep them updated. They are not used anywhere in contracts anyways. Should add `height` to show the indexed height of the info. Should add boolean `jailed` to indicate whether the fp is jailed
2. [finality provider info](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/9fff11cf8c31523fbe3df75491af9f81ad90064f/contracts/btc-staking/src/contract.rs#L83). Should add `height` to show the indexe height of the info. Should be moved to the finality module

~3. [finality provider by power](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/9fff11cf8c31523fbe3df75491af9f81ad90064f/contracts/btc-staking/src/contract.rs#L86) can be removed as there already exists such query in the finality contract~
~4. [Activated Height](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/9fff11cf8c31523fbe3df75491af9f81ad90064f/contracts/btc-staking/src/contract.rs#L89) should be moved to finality contracts and align with the concept in Babylon Genesis' [activated height](https://github.com/babylonlabs-io/babylon/blob/49eabfd83a1bcc0a9bae99ef4f6f452d529966fb/proto/babylon/finality/v1/query.proto#L38)~

gitferry:
#276 fixed (3). In light of [discussions](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/276#discussion_r2246697982), (1) don't need to add `height` and `jailed` anymore. (2) `finality_provider_info` can be removed.

(4) will be tacked in #284

PR #276, opened 2025-07-31
finality: refactor voting power table in BTC finality contract, and fix finality signature handling/tallying

### LINKED ISSUE #156: Voting power and finality sig issues
There are few issues noticed in the contracts

1. Finality sigs are being accepted even when there is no active FP set. This is partly happening b/c the voting power in contract storage is bit wrong, when delegation is active we increase VP and only reduce later when its unbonded (manually or natural expiration). As per implementation in Babylon - there should be no VP until pub rand is timestamped but step is missing in contracts
2. When active FP set is computed (in begin blocker of finality contract). I noticed we still update the voting power table (empty vector), then later when block tallying happens (in end blocker of finality contract), it matches to the wrong switch case. Basically blocks which have no active FP set should not be tallied, they should be skipped. We should not save empty FP set [here](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/d51d97a42d9aa394b374d513f35367e4ed7e7a2e/contracts/btc-finality/src/finality.rs#L729), doing so cause the tally to get stuck [here](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/d51d97a42d9aa394b374d513f35367e4ed7e7a2e/contracts/btc-finality/src/finality.rs#L518) b/c `Some(fp_set) `will be true

Some more context from e2e tests

Staking activated at BSN height 467
```
root@c7d70c51e69f:/ibcsim-bcd# bcd query wasm contract-state smart bbnc17p9rzwnnfxcjp32un9ug7yhhzgtkhvl9jfksztgw5uh69wac2pgssg3nft '{"activated_height":{}}'
data:
  height: 467
```

No active FPs at this height, b/c pub randomness not timestamped yet
```
root@c7d70c51e69f:/ibcsim-bcd# bcd query wasm contract-state smart bbnc1ghd753shjuwexxywmgs4xz7x2q732vcnkm6h2pyv9s6ah3hylvrqv7u2f2 '{"active_finality_providers":{"height": 467}}'
data:
  active_finality_providers: []
```

But the finality sig was accepted at this height
```
root@c7d70c51e69f:/ibcsim-bcd# bcd query wasm contract-state smart bbnc1ghd753shjuwexxywmgs4xz7x2q732vcnkm6h2pyv9s6ah3hylvrqv7u2f2 '{"finality_signature":{"height": 476, "btc_pk_hex": "8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b"}}'
data:
  signature:
  - 62
  - 107
  - 233
  - 194
  - 7
  - 62
  - 6
  - 182
...
```

You can see the pub rand height
```
root@c7d70c51e69f:/ibcsim-bcd# bcd query wasm contract-state smart bbnc1ghd753shjuwexxywmgs4xz7x2q732vcnkm6h2pyv9s6ah3hylvrqv7u2f2 '{"pub_rand_commit":{"btc_pk_hex": "8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b", "reverse": true }}' --output json | jq '[.data[] | del(.commitment)]'
[
  {
    "start_height": 2325,
    "num_pub_rand": 1000,
    "height": 1326
  },
  {
    "start_height": 1325,
    "num_pub_rand": 1000,
    "height": 330
  },
  {
    "start_height": 325,
    "num_pub_rand": 1000,
    "height": 323
  }
]
```

Notice the log here from finality contract begin blocker - `(finalized_height: 304 >= commit_height: 323)`
```
gusin@Gurjots-MacBook-Pro babylon-contract % docker logs ibcsim-bcd | grep -B1 -A20 -E 'compute_active_finality_providers called at height: 467([^0-9]|$)'

5:41PM WRN cannot get tx contracts from context module=x/wasm
[2025-06-16T17:41:58.157342013Z]: DEBUG: compute_active_finality_providers called at height: 467, max_active_fps: 100 (gas remaining: 18446744072747525665)
[2025-06-16T17:41:58.157375013Z]: DEBUG: Loaded config, staking contract: bbnc17p9rzwnnfxcjp32un9ug7yhhzgtkhvl9jfksztgw5uh69wac2pgssg3nft (gas remaining: 18446744072504599055)
5:41PM WRN cannot get tx contracts from context module=x/wasm
[2025-06-16T17:41:58.15763993Z]: DEBUG: Last finalized height: 304 (gas remaining: 18446744063131176295)
5:41PM WRN cannot get tx contracts from context module=x/wasm
[2025-06-16T17:41:58.157965971Z]: DEBUG: Initial batch from staking contract: 1 FPs (gas remaining: 18446744053826811135)
[2025-06-16T17:41:58.157977763Z]: DEBUG: Initial FP[0]: btc_pk_hex=8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b, power=1000000 (gas remaining: 18446744053825842070)
[2025-06-16T17:41:58.15798293Z]: DEBUG: Filtering FP: btc_pk_hex=8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b, power=1000000 (gas remaining: 18446744053824924410)
[2025-06-16T17:41:58.158005805Z]: DEBUG: has_timestamped_pub_rand_commit_for_height called - FP: 8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b, height: 467, last_finalized_height: Some(304) (gas remaining: 18446744053639761355)
[2025-06-16T17:41:58.158069721Z]: DEBUG: Found pub rand commit - start_height: 325, num_pub_rand: 1000, commit_height: 323 (gas remaining: 18446744053363707395)
[2025-06-16T17:41:58.158074346Z]: DEBUG: Using provided finalized height: 304 (gas remaining: 18446744053363183610)
[2025-06-16T17:41:58.158079221Z]: DEBUG: Timestamped check result: false (finalized_height: 304 >= commit_height: 323) (gas remaining: 18446744053362311950)
[2025-06-16T17:41:58.158083721Z]: DEBUG: FP 8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b filtered out: no timestamped public randomness for height 467 (gas remaining: 18446744053361484565)
[2025-06-16T17:41:58.158087971Z]: DEBUG: Filtered batch: 0 FPs passed filters (gas remaining: 18446744053360808865)
5:41PM WRN cannot get tx contracts from context module=x/wasm
[2025-06-16T17:41:58.158343471Z]: DEBUG: Next batch from staking contract: 0 FPs (gas remaining: 18446744044400485450)
[2025-06-16T17:41:58.158351846Z]: DEBUG: Final finality_providers count: 0, total_power: 0 (gas remaining: 18446744044399793420)
[2025-06-16T17:41:58.158379263Z]: DEBUG: Saving 0 active FPs to FP_SET at height 467 (gas remaining: 18446744044010814295)
5:41PM DBG response of sudo call [123 34 98 101 103 105 110 95 98 108 111 99 107 34 58 123 34 104 97 115 104 95 104 101 120 34 58 34 102 102 53 51 57 48 57 57 54 98 55 54 49 49 102 56 50 100 50 101 51 54 98 97 57 55 101 98 98 56 48 101 99 97 100 98 57 100 97 55 53 56 57 52 57 49 97 98 56 52 100 48 48 49 51 53 98 99 54 56 52 100 52 98 34 44 34 97 112 112 95 104 97 115 104 95 104 101 120 34 58 34 97 102 51 49 51 50 57 55 49 50 100 51 50 55 97 102 50 53 100 100 97 57 51 97 99 101 101 55 51 51 98 56 53 50 97 102 53 50 51 55 98 100 51 48 101 97 101 102 53 54 98 97 56 51 99 99 100 55 51 57 48 52 54 55 34 125 125] to contract bbnc1ghd753shjuwexxywmgs4xz7x2q732vcnkm6h2pyv9s6ah3hylvrqv7u2f2: [] module=x/babylon
```

Notice the log here that finality sig was accepted even though no active FP at this height
```
gusin@Gurjots-MacBook-Pro babylon-contract % docker logs ibcsim-bcd | grep -A10 "DEBUG: handle_finality_signature called - FP: 8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b, height: 467"
[2025-06-16T17:42:01.444552125Z]: DEBUG: handle_finality_signature called - FP: 8ec602741ac99d72ce17a0f39db59988a899362e61203f507d4533a184873a6b, height: 467, current_block_height: 470 (gas remaining: 18446744073682807945)
[2025-06-16T17:42:01.444567917Z]: DEBUG: pub_rand length: 32, signature length: 32, block_app_hash: af31329712d327af25dda93acee733b852af5237bd30eaef56ba83ccd7390467 (gas remaining: 18446744073680460375)
[2025-06-16T17:42:01.444591292Z]: DEBUG: Loaded staking contract address: bbnc17p9rzwnnfxcjp32un9ug7yhhzgtkhvl9jfksztgw5uh69wac2pgssg3nft (gas remaining: 18446744073437405540)
5:42PM WRN cannot get tx contracts from context module=x/wasm
[2025-06-16T17:42:01.445064375Z]: DEBUG: Found finality provider - slashed_height: 0 (gas remaining: 18446744063966585795)
[2025-06-16T17:42:01.445072Z]: DEBUG: FP is not slashed, continuing (gas remaining: 18446744063966322255)
[2025-06-16T17:42:01.44507525Z]: DEBUG: Querying FP voting power for height 467 (gas remaining: 18446744063965816180)
5:42PM WRN cannot get tx contracts from context module=x/wasm
[2025-06-16T17:42:01.445287125Z]: DEBUG: FP voting power at height 467: 1000000 (gas remaining: 18446744054770109950)
[2025-06-16T17:42:01.445290709Z]: DEBUG: Height and signature checks passed (gas remaining: 18446744054769841465)
[2025-06-16T17:42:01.445310292Z]: DEBUG: No existing signature found for this FP and height (gas remaining: 18446744054579583995)
```

```
gusin@Gurjots-MacBook-Pro babylon-contract % docker logs ibcsim-bcd | grep -A10 "DEBUG: tally_blocks called at height 467"
[2025-06-16T17:41:58.159116221Z]: DEBUG: tally_blocks called at height 467 with activated_height 467 (gas remaining: 18446744063197730455)
[2025-06-16T17:41:58.159129138Z]: DEBUG: tally_blocks will process blocks from 467 to 467 (gas remaining: 18446744063038330075)
[2025-06-16T17:41:58.159133263Z]: DEBUG: Processing block height 467 (gas remaining: 18446744063037787085)
[2025-06-16T17:41:58.159161971Z]: DEBUG: Block 467 - has_fp_set: true, is_finalized: false (gas remaining: 18446744062641788100)
[2025-06-16T17:41:58.15916793Z]: DEBUG: Block 467 has 0 active FPs, attempting to tally votes (gas remaining: 18446744062641040640)
[2025-06-16T17:41:58.159208138Z]: DEBUG: Block 467 received 0 signatures (gas remaining: 18446744062634550305)
[2025-06-16T17:41:58.15921368Z]: DEBUG: Block 467 did not achieve >2/3 votes, stopping finalization (gas remaining: 18446744062633848730)
[2025-06-16T17:41:58.159217888Z]: DEBUG: tally_blocks completed - finalized 0 blocks (gas remaining: 18446744062633217765)
[2025-06-16T17:41:58.159221138Z]: DEBUG: No blocks finalized, no rewards to mint (gas remaining: 18446744062632909605)
```

SebastianElvis:
Resolves #156 

This PR fixes the voting power table bugs. This includes

- [x] replacing voting power with total active sat in BTC staking contract. Now BTC staking contract is only tracking total active sat and slashing status for each FP
- [x] refactoring `compute_active_finality_providers` in BTC finality contract such that it maintains the voting power table internally
- [x] `compute_active_finality_providers` now excludes FPs that do not have BTC timestamps, are slashed, are jailed, or have 0 total active sat
- [x] BTC finality contract now uses its internal voting power table for handling finality signatures and tallying
- [x] adding a query for FP voting power in BTC finality contract
- [x] adding unit tests in BTC staking contract

### DIFF
--- a/contracts/btc-finality/src/finality.rs
+++ b/contracts/btc-finality/src/finality.rs
@@ -252,7 +252,7 @@ pub fn handle_finality_signature(
     //     this means Babylon will lose safety under an adaptive adversary corrupting even 1
     //     finality provider. It can simply corrupt a new finality provider and equivocate a
     //     historical block over and over again, making a previous block not finalisable forever
-    if fp.slashed_height > 0 && fp.slashed_height < height {
+    if fp.slashed_height > 0 {
         return Err(ContractError::FinalityProviderAlreadySlashed(
             fp_btc_pk_hex.to_string(),
         ));
--- a/contracts/btc-finality/src/finality.rs
+++ b/contracts/btc-finality/src/finality.rs
@@ -749,11 +749,9 @@ pub fn compute_active_finality_providers(
 
     // Online FPs verification
     // Store starting heights of fps entering the active set
-    let old_fps = get_power_table_at_height(deps.storage, env.block.height - 1)?
-        .keys()
-        .cloned()
-        .collect();
-    let cur_fps: HashSet<String> = fp_power_table.keys().cloned().collect();
+    let old_power_table = get_power_table_at_height(deps.storage, env.block.height - 1)?;
+    let old_fps = old_power_table.keys().collect();
+    let cur_fps: HashSet<_> = fp_power_table.keys().collect();
     let new_fps = cur_fps.difference(&old_fps);
     for fp in new_fps {
         // Active since the next block. Only save if not already set
@@ -798,9 +796,7 @@ pub fn compute_active_finality_providers(
     Ok(())
 }
 
-/// Query BTC staking contract for finality providers by total active sats.
-///
-/// This is used to get the finality providers with the highest total active sats.
+/// Queries the BTC staking contract for finality providers ordered by total active sats.
 pub fn query_fps_by_total_active_sats(
     staking_addr: &Addr,
     querier: &QuerierWrapper,
--- a/contracts/btc-finality/src/finality.rs
+++ b/contracts/btc-finality/src/finality.rs
@@ -2,7 +2,8 @@ use crate::contract::encode_smart_query;
 use crate::error::ContractError;
 use crate::state::config::{ADMIN, CONFIG, PARAMS};
 use crate::state::finality::{
-    BLOCKS, EVIDENCES, FP_BLOCK_SIGNER, FP_SET, FP_START_HEIGHT, JAIL, NEXT_HEIGHT, SIGNATURES,
+    get_power_table_at_height, BLOCKS, EVIDENCES, FP_BLOCK_SIGNER, FP_POWER_TABLE, FP_START_HEIGHT,
+    JAIL, NEXT_HEIGHT, SIGNATURES,
 };
 use crate::state::public_randomness::{
     get_last_finalized_height, get_last_pub_rand_commit,
@@ -13,7 +14,7 @@ use babylon_apis::btc_staking_api::FinalityProvider;
 use babylon_apis::finality_api::{Evidence, IndexedBlock, PubRandCommit};
 use babylon_apis::to_canonical_addr;
 use babylon_merkle::Proof;
-use btc_staking::msg::{FinalityProviderInfo, FinalityProvidersByPowerResponse};
+use btc_staking::msg::{FinalityProviderInfo, FinalityProvidersByTotalActiveSatsResponse};
 use cosmwasm_std::Order::Ascending;
 use cosmwasm_std::{
     to_json_binary, Addr, DepsMut, Env, Event, MessageInfo, QuerierWrapper, Response, StdResult,
@@ -22,7 +23,7 @@ use cosmwasm_std::{
 use k256::schnorr::{signature::Verifier, Signature, VerifyingKey};
 use k256::sha2::{Digest, Sha256};
 use std::cmp::max;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 // The maximum number of blocks into the future
 // that a public randomness commitment start height can target. This limit prevents abuse by capping
@@ -258,17 +259,8 @@ pub fn handle_finality_signature(
     }
 
     // Ensure the finality provider has voting power at this height
-    let fp: FinalityProviderInfo = deps
-        .querier
-        .query_wasm_smart(
-            staking_addr.clone(),
-            &btc_staking::msg::QueryMsg::FinalityProviderInfo {
-                btc_pk_hex: fp_btc_pk_hex.to_string(),
-                height: Some(height),
-            },
-        )
-        .map_err(|_| ContractError::NoVotingPower(fp_btc_pk_hex.to_string(), height))?;
-    if fp.power == 0 {
+    let fp_power = FP_POWER_TABLE.load(deps.storage, (height, fp_btc_pk_hex))?;
+    if fp_power == 0 {
         return Err(ContractError::NoVotingPower(
             fp_btc_pk_hex.to_string(),
             height,
@@ -614,16 +606,17 @@ pub fn tally_blocks(
     for h in start_height..=env.block.height {
         let mut indexed_block = BLOCKS.load(deps.storage, h)?;
         // Get the finality provider set of this block
-        let fp_set = FP_SET.may_load(deps.storage, h)?;
+        let fp_power_table = get_power_table_at_height(deps.storage, h)?;
+        let has_fp = !fp_power_table.is_empty();
 
-        match (fp_set, indexed_block.finalized) {
-            (Some(fp_set), false) => {
+        match (has_fp, indexed_block.finalized) {
+            (true, false) => {
                 // Has finality providers, non-finalised: tally and try to finalise the block
                 let voter_btc_pks = SIGNATURES
                     .prefix(indexed_block.height)
                     .keys(deps.storage, None, None, Ascending)
                     .collect::<StdResult<Vec<_>>>()?;
-                if tally(&fp_set, &voter_btc_pks) {
+                if tally(&fp_power_table, &voter_btc_pks) {
                     // If this block gets >2/3 votes, finalise it
                     let ev = finalize_block(deps.storage, &mut indexed_block, &voter_btc_pks)?;
                     events.push(ev);
@@ -633,20 +626,20 @@ pub fn tally_blocks(
                     break;
                 }
             }
-            (None, false) => {
+            (false, false) => {
                 // Does not have finality providers, non-finalised: not finalisable,
                 // Increment the next height to finalise and continue
                 NEXT_HEIGHT.save(deps.storage, &(indexed_block.height + 1))?;
                 continue;
             }
-            (Some(_), true) => {
+            (true, true) => {
                 // Has finality providers and the block is finalised.
                 // This can only be a programming error
                 return Err(ContractError::FinalisedBlockWithFinalityProviderSet(
                     indexed_block.height,
                 ));
             }
-            (None, true) => {
+            (false, true) => {
                 // Does not have finality providers, finalised: impossible to happen
                 return Err(ContractError::FinalisedBlockWithoutFinalityProviderSet(
                     indexed_block.height,
@@ -659,14 +652,14 @@ pub fn tally_blocks(
 }
 
 /// Checks whether a block with the given finality provider set and votes reaches a quorum or not.
-fn tally(fp_set: &[FinalityProviderInfo], voters: &[String]) -> bool {
+fn tally(fp_power_table: &HashMap<String, u64>, voters: &[String]) -> bool {
     let voters: HashSet<String> = voters.iter().cloned().collect();
     let mut total_power = 0;
     let mut voted_power = 0;
-    for fp_info in fp_set {
-        total_power += fp_info.power;
-        if voters.contains(&fp_info.btc_pk_hex) {
-            voted_power += fp_info.power;
+    for (fp_btc_pk_hex, power) in fp_power_table {
+        total_power += power;
+        if voters.contains(fp_btc_pk_hex) {
+            voted_power += power;
         }
     }
     voted_power * 3 > total_power * 2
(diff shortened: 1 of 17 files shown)


gitferry (commented):
Tbh I feel this PR has mixed with a bunch of refactoring (e.g., partly fixed the issue https://github.com/babylonlabs-io/cosmos-bsn-contracts/issues/268) and I failed to see where the fix is and how does the fix work. Could you point me there?

contracts/btc-staking/src/msg.rs
```diff
@@ -111,9 +111,10 @@ pub struct FinalityProviderInfo {
     /// Bitcoin secp256k1 PK of this finality provider.
     /// The PK follows encoding in BIP-340 spec in hex format
     pub btc_pk_hex: String,
-    /// Aggregated power of this finality provider.
-    /// The power is calculated based on the amount of BTC delegated to this finality provider
-    pub power: u64,
+    /// Total active sats delegated to this finality provider
+    pub total_active_sats: u64,
+    /// Whether this finality provider is slashed
+    pub slashed: bool,
```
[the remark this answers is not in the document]
SebastianElvis: `FinalityProvider` is the object for storing the FP metadata, i.e., same as the one in Babylon. It is better aligned with Babylon's FP definition.

meanwhile, this object aims to track the factors that determine the voting power of FPs, in particular the total active delegations and slashed. BTC finality contract will consume the object and determine the FP's voting power based on these factors, together with timestamped pub rand and jailing.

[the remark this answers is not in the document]
SebastianElvis: Also, this object is in BTC staking contract that does not have the concept of jailing, so we could not have `jailed` flag here

SebastianElvis:
> Tbh I feel this PR has mixed with a bunch of refactoring (e.g., partly fixed the issue #268) and I failed to see where the fix is and how does the fix work. Could you point me there?

I think #156 gives more context on the issue. To reiterate

- The bug is that BTC finality contract accepts finality signatures from FPs who even do not have timestamped pub rand
- The fundamental reason of this bug is twofold
  - BTC staking contract maintains voting power, which does not (and cannot) take into account timestamped pub rand or jailing
  - BTC finality contract is using the voting power table in BTC staking contract

Thus to fix the issue completely, we need to 

- refactor the voting power table to clearly distinguish between 
  - total active sats in BTC staking contract
  - voting power table in BTC finality contract
- remove the concept of voting power in BTC staking contract completely
- use voting power table for all actions in BTC finality contract, including handling finality signatures and tallying

All these issues have to be fixed within a single PR imo. I revised the PR title to make this more clear

contracts/btc-finality/src/finality.rs
```diff
@@ -252,29 +252,14 @@ pub fn handle_finality_signature(
...
-            },
-        )
-        .map_err(|_| ContractError::NoVotingPower(fp_btc_pk_hex.to_string(), height))?;
-    if fp.power == 0 {
-        return Err(ContractError::NoVotingPower(
-            fp_btc_pk_hex.to_string(),
-            height,
-        ));
-    }
+    ensure_fp_has_power(deps.storage, height, fp_btc_pk_hex)?;
```
SebastianElvis, line 262: @gitferry After this change, the finality sig handler is using voting power table in BTC finality contract, not BTC staking contract. It filters out FPs without timestamped pub rand, thus fixes #156

contracts/btc-finality/src/finality.rs
```diff
@@ -615,16 +598,17 @@ pub fn tally_blocks(
...
         // Get the finality provider set of this block
-        let fp_set = FP_SET.may_load(deps.storage, h)?;
+        let fp_power_table = get_power_table_at_height(deps.storage, h)?;
+        let has_fp = !fp_power_table.is_empty();
 
-        match (fp_set, indexed_block.finalized) {
-            (Some(fp_set), false) => {
+        match (has_fp, indexed_block.finalized) {
+            (true, false) => {
                 // Has finality providers, non-finalised: tally and try to finalise the block
```
SebastianElvis, line 606: @gitferry After this change, tallying is using the voting power table in BTC finality contract. This was also not correct because the "voting power" in BTC staking contract is not correct and does not account for timestamped pub rand or jailing

contracts/btc-staking/src/msg.rs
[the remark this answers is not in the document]
gitferry: The btc-staking contract now has `FinalityProvider` for meta data and `FinalityProviderInfo` for run-time states. This is not aligned with Babylon Genesis which only has `FinalityProvider` to carry both meta data and states. I also feel it's confusing and not necessary to have two endpoints for finality providers. We can merge `FinalityProviderInfo` to either `FinalityProvider` or `FinalityProviderPower` in the finality contract. Either way we can fix it in another pr when tackling #268. Wdyt?

gitferry (approved):
Ah now I get the issue is that we used to use the total_active_sats stored in the btc-staking contracts as voting power which is definitely not true. The fix lgtm and aligned with Babylon now. My concern is shouldn't we have tests to uncover this case?

contracts/btc-staking/src/msg.rs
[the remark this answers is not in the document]
SebastianElvis: > This is not aligned with Babylon Genesis which only has FinalityProvider to carry both meta data and states

This is not fully aligned with Babylon in the sense that tracking total sats is in finality module not in BTC staking module, but not in the sense that "FinalityProvider to carry both meta data and states".

So in this implementation, `FinalityProviderInfo` is mirroring the `FinalityProviderDistInfo` struct in [Babylon](https://github.com/babylonlabs-io/babylon/blob/1bd8f3930416a700a2b846beedd5d96d968a25c8/proto/babylon/finality/v1/finality.proto#L25-L55

The problem of this `FinalityProviderDistInfo` in Babylon is that, it tracks states across two different modules

- jailed / timestamped pub rand are states in finality module
- total active sats and slashed are states in BTC staking module

This is a limitation not a feature. It also adds quite some complexity in maintaining these states, despite we are still iterating over all FPs upon each BeginBlock in Babylon. The current implementation in the contracts avoids this particular problem

> . I also feel it's confusing and not necessary to have two endpoints for finality providers.

I believe Babylon also has endpoints for FPs in both BTC staking module and finality module, with different focus. One thing that we can consider though, is to remove the `FinalityProviderInfo` [interface](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/a2dfe6c8d3e7bdb4ebf5fbd3fb79074dbb9c6e61/contracts/btc-staking/src/msg.rs#L68-L76)  in the contract. It is not used anymore. But can be done in separate PRs

contracts/btc-finality/src/finality.rs
```diff
@@ -707,53 +691,67 @@ pub fn compute_active_finality_providers(
...
             .into_iter()
             .filter(|fp| {
-                // Filter out FPs with no voting power
-                if fp.power == 0 {
+                // Filter out FPs with no active sats
+                if fp.total_active_sats == 0 {
+                    return false;
+                }
+                // Filter out slashed FPs
+                if fp.slashed {
```
maurolacy, line 709: This is a bug fix. Good.

contracts/btc-finality/src/finality.rs
```diff
@@ -707,53 +691,67 @@ pub fn compute_active_finality_providers(
...
+                if JAIL
+                    .may_load(deps.storage, &fp.btc_pk_hex)
                     .unwrap_or(Some(JAIL_FOREVER))
-                    .is_none()
+                    .is_some()
+                {
+                    return false;
+                }
                 // Filter out FPs that don't have timestamped public randomness
-                && has_timestamped_pub_rand_commit_for_height(&deps.as_ref(), &fp.btc_pk_hex, env.block.height, Some(last_finalized_height))
```
maurolacy, line 730: Here's the original check for timestamped pub rand commit for height.

Correct me if I'm wrong, but even if a FP was allowed to submit a sig without timestamped pub rand, he would have been excluded from the tallying here.

So, besides some renaming for clarity (which is good) and a bug fix related to filtering out slashed FPs, this PR looks more like an alignment with Babylon than anything else to me (which again, is good).

contracts/btc-finality/src/finality.rs
```diff
@@ -787,37 +785,45 @@ pub fn compute_active_finality_providers(
     })?;
 
     // Save the new set of active finality providers
-    // TODO: Purge old (height - finality depth) FP_SET entries to avoid bloating the storage (#124)
-    FP_SET.save(deps.storage, env.block.height, &finality_providers)?;
+    for (fp_btc_pk_hex, power) in fp_power_table {
```
maurolacy, line 788: It's more efficient to have all active FPs in a set; basically because they are being loaded all at the same time, and saved all at the same time.
The only reason for using a map is, if you want to query the power of a given (active) FP (at a given height).

contracts/btc-finality/src/finality.rs
maurolacy: This is the important fix.

This could have been done by checking that the FP is in the active set (i.e. in `FP_SET`) at the given height).

I agree this impl is better, as loading the entire set at each finality sig submission would be inefficient.

maurolacy (commented):
Looks good to me in the renaming part, but I would have preferred to keep the active FPs in a set, not a hash map.

A way to do that would have been, to add / replicate the timestamped pub rand check for the FP, directly in / during sig submission.

contracts/btc-staking/src/msg.rs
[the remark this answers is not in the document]
gitferry: > I believe Babylon also has endpoints for FPs in both BTC staking module and finality module, with different focus. One thing that we can consider though, is to remove the FinalityProviderInfo [interface](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/a2dfe6c8d3e7bdb4ebf5fbd3fb79074dbb9c6e61/contracts/btc-staking/src/msg.rs#L68-L76) in the contract. It is not used anymore. But can be done in separate PRs

Agreed on this

contracts/btc-finality/src/finality.rs
SebastianElvis: So if the FP has no timestamped pub rand at this height, it won't be included in the voting power table and the finality sig will be rejected due to zero voting power

PR #279, opened 2025-08-01
chore: clean up BTC finality contract parameters and merge them into `Config`

### LINKED ISSUE #103: Flatten instantiation config / params

### LINKED ISSUE #218: remove unnecessary parameters in contracts
The parameters should also be versioned

SebastianElvis:
Resolves #218 
Resolves #103

This PR cleans up BTC finality contract parameters and merge them into `Config` object.


contracts/btc-finality/src/contract.rs
```diff
@@ -268,8 +244,7 @@ fn handle_end_block(
     }
 
     // On an epoch boundary, send rewards for distribution to Babylon Genesis
-    let params = PARAMS.load(deps.storage)?;
-    if env.block.height > 0 && env.block.height % params.reward_interval == 0 {
+    if env.block.height > 0 && env.block.height % cfg.reward_interval == 0 {
```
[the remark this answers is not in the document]
SebastianElvis: false positive, `cfg` is initialised at L234

contracts/btc-finality/src/contract.rs
```diff
@@ -361,7 +336,11 @@ pub(crate) mod tests {
 
         // Create an InstantiateMsg with admin set to None
         let msg = InstantiateMsg {
```
liuchengxu, line 338: This can be simply `InstantiateMsg::default()`.

contracts/babylon/src/multitest.rs
```diff
@@ -130,16 +130,12 @@ mod instantiation {
...
     fn instantiate_finality_msg_works() {
-        // Params setting is an all-or-nothing operation, i.e. all the params have to be set
-        let params = btc_finality::state::config::Params {
-            reward_interval: 10,
-            max_active_finality_providers: 5,
-            min_pub_rand: 2,
-            missed_blocks_window: 100,
-            jail_duration: 3600,
-        };
         let finality_instantiation_msg = btc_finality::msg::InstantiateMsg {
```
liuchengxu, line 133:
```suggestion
        let finality_instantiation_msg = btc_finality::msg::InstantiateMsg::default();
```

contracts/btc-finality/src/contract.rs
```diff
@@ -385,7 +364,11 @@ pub(crate) mod tests {
 
         // Create an InstantiateMsg with admin set to Some(INIT_ADMIN.into())
         let msg = InstantiateMsg {
```
liuchengxu, line 366:
```suggestion
        let msg = InstantiateMsg {
            admin: Some(init_admin.to_string()), // Admin provided
            ..Default::default()
        };
```

contracts/btc-finality/src/contract.rs
```diff
@@ -533,7 +524,11 @@ pub(crate) mod tests {
 
         // Create an InstantiateMsg with admin set to Some(INIT_ADMIN.into())
         let instantiate_msg = InstantiateMsg {
```
liuchengxu, line 526: Same as above.

contracts/btc-finality/tests/integration.rs
```diff
@@ -29,7 +29,11 @@ fn instantiate_works() {
     let mut deps = mock_instance(WASM, &[]);
 
     let msg = InstantiateMsg {
```
liuchengxu, line 31:
```suggestion
    let msg = InstantiateMsg::default();
```

contracts/btc-finality/src/multitest/suite.rs
```diff
@@ -153,7 +149,11 @@ impl SuiteBuilder {
                     btc_finality_code_id: Some(btc_finality_code_id),
                     btc_finality_msg: Some(
                         to_json_binary(&InstantiateMsg {
```
liuchengxu, line 151: Same as above.

liuchengxu (approved):

contracts/btc-finality/src/msg.rs
```diff
@@ -11,8 +10,12 @@ use {
 #[cw_serde]
 #[derive(Default)]
 pub struct InstantiateMsg {
-    pub params: Option<Params>,
     pub admin: Option<String>,
+    pub max_active_finality_providers: Option<u32>,
+    pub min_pub_rand: Option<u64>,
+    pub reward_interval: Option<u64>,
+    pub missed_blocks_window: Option<u64>,
+    pub jail_duration: Option<u64>,
```
maurolacy, line 18: 👍🏼

maurolacy (commented):
LGTM. Thanks for doing this.

PR #286, opened 2025-08-03
fix: ICS-20 transfer memo format

maurolacy:
A fix for the FP BTC PK format.

### DIFF
--- a/contracts/babylon/src/msg/ibc.rs
+++ b/contracts/babylon/src/msg/ibc.rs
@@ -1,7 +1,6 @@
 use ibc_proto::ibc::core::channel::v1::{acknowledgement::Response, Acknowledgement};
 
 use cosmwasm_schema::cw_serde;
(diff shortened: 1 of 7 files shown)


contracts/babylon/src/msg/ibc.rs
```diff
@@ -48,6 +47,7 @@ pub struct BsnRewards {
 
 #[cw_serde]
 pub struct FpRatio {
-    pub btc_pk: Binary,
+    // FP BTC public key, encoded in hex
+    pub btc_pk: String,
```
liuchengxu, line 51: We may consider a unit type wrapper or simply an alias, something like `BtcPubKey` for the BTC public key encoded in hex string, so that it's super clear what it's from the type itself. Can be done in the future.

maurolacy: Usually there's a convention that a hex field name ends in `_hex`.

PR #288, opened 2025-08-04
finality: add `should_accept_sig_for_height()` check

### LINKED ISSUE #263: Missalignment in finality votes verification

liuchengxu:
Close #263

### DIFF
--- a/contracts/btc-finality/src/finality.rs
+++ b/contracts/btc-finality/src/finality.rs
@@ -352,12 +352,57 @@ impl MsgAddFinalitySig {
     }
 }
 
+/// Determines whether the finality signature should be accepted for a given block height.
+///
+/// Babylon Genesis and BSN use different rules for accepting signatures based on how
+/// their Bitcoin timestamping protocols operate:
+///
+/// - **Babylon Genesis** timestamps *epochs*, and each epoch includes a fixed number of blocks.
+///   Therefore, each `pub_rand` commit is tagged by an *epoch number*, and is considered timestamped
+///   if the corresponding epoch has been timestamped on Bitcoin.
+///
+/// - **BSN**, on the other hand, timestamps *individual blocks*. It relies on Babylon Genesis
+///   to perform the actual Bitcoin timestamping by submitting BSN headers to Babylon via IBC.
+///   As a result, each `pub_rand` commit is tagged with a *BSN block height*, and is considered
+///   timestamped if the highest Bitcoin-finalized BSN height is greater than or equal to that value.
+///
+/// This difference leads to separate implementations of `should_accept_sig_for_height` in both systems.
+///
+/// https://github.com/babylonlabs-io/babylon/blob/49972e2d3e35caf0a685c37e1f745c47b75bfc69/x/finality/keeper/msg_server.go#L245
+fn should_accept_sig_for_height(
+    storage: &dyn Storage,
+    indexed_block: &IndexedBlock,
+) -> cosmwasm_std::StdResult<bool> {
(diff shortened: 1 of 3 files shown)


contracts/btc-finality/src/finality.rs
```diff
@@ -352,13 +352,58 @@ impl MsgAddFinalitySig {
...
+    indexed_block: &IndexedBlock,
+) -> cosmwasm_std::StdResult<bool> {
+    let Some(next_height) = NEXT_HEIGHT.may_load(storage)? else {
+        // Accept finality sig if there is no any finalized block yet.
+        return Ok(true);
+    };
+
+    let last_finalized_height = next_height.saturating_sub(1);
+
+    let timestamped = last_finalized_height >= indexed_block.height;
```
SebastianElvis, line 383: This is not equivalent to Babylon implementation. 

There are two notions of finalisation in Babylon: 

- BTC timestamp finalised: A block is BTC timestamped
- BTC staking finalised: A block receives quorum number of finality signatures backed by BTC stake

Here `last_finalized_height` is the last BTC staking finalised height, not the last BTC timestamping finalised height. Thus the result of variable `timestamped` is not correct

SebastianElvis: Instead, what we need to do is to

- Query the Babylon contract on the last BTC timestamp with non-empty consumer header
- Get the height of the consumer header as `last_finalized_height`, i.e., the last BTC timestamping finalised height
- Derive `timestamped = last_finalized_height >= indexed_block.height`

However I don't think this is straightforward due to the cross-contract query. Imo this is an optimisation that is nice to have, not must-have. If we have other urgent task (which I believe so) maybe we can leave it in a Github issue for now, wdyt @gitferry

liuchengxu: So there is a difference between the btc_timestamp_finalized and btc_staking_finalized, is it an acceptable misalignment if we use btc_staking_finalized while Babylon Genesis uses btc_timestamp_finalized? I guess the btc_staking_finalization is usually the faster one, right?

liuchengxu: Created #289 and clarified the impl in the code.

liuchengxu: Correct me if I'm wrong, the two finalization mechanisms run in parallel, the BTC staking finalization is faster but weaker. As mentioned by Gai, this is an optimization by Babylon Genesis to avoid redundantly voting for the blocks that have been timestamped.

That said, I agree with Gai that we may keep this optimization as a TODO in this PR instead of closing it directly.

PR #290, opened 2025-08-04
finality: add missing finality queries

### LINKED ISSUE #269: Misalignment in the finality contract queries
1. Miss votes at height https://github.com/babylonlabs-io/babylon/blob/49eabfd83a1bcc0a9bae99ef4f6f452d529966fb/proto/babylon/finality/v1/query.proto#L66
2. Miss signing info(s) https://github.com/babylonlabs-io/babylon/blob/49eabfd83a1bcc0a9bae99ef4f6f452d529966fb/proto/babylon/finality/v1/query.proto#L81. Given that the jailing mechanism between contract and go module are different, we can return the jailing info relevent to the contract, e.g., the last signed height https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/9fff11cf8c31523fbe3df75491af9f81ad90064f/contracts/btc-finality/src/state/finality.rs#L24

liuchengxu:
I'm not confident that the `signing_info()` query semantically aligns with the existing ones. Please take a closer look.

Close #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Improvements
 
+- [#290](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/290) finality: add missing queries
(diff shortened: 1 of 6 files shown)


gitferry (approved):
Lgtm!

PR #291, opened 2025-08-04
rewards: improve rewards dist

### LINKED ISSUE #277: Misalignment in rewards distribution
### Current Behavior

Rewards are accumulated per BeginBlock based on each finality provider's (FP) voting power from the previous block. Accumulated rewards are distributed at a fixed block interval. This approach does not fully align with the Babylon Genesis design, where rewards should reflect voting participation,

### Problem

The current implementation may unfairly reward FPs that fail to submit finality signatures within a reasonable timeframe, undermining the security model.

### Proposed Solution

Introduce a `finality_sig_timeout` parameter to enforce a deadline for finality signatures.  Modify rewards logic as follows:

#### Per BeginBlock (Height H):

- Handle rewards for `height H - finality_sig_timeout`. This ensures FPs have sufficient time to submit signatures (ref: [Babylon ABCI logic](https://github.com/babylonlabs-io/babylon/blob/1a05ecd8dfc69691b6c17637ef520ce9ec302113/x/finality/abci.go#L36)).
- Calculate rewards for the target height:
`current_rewards = total_balance_of_finality - total_accumulated_rewards ` where `total_balance_of_finality` is the balance of the finality contract that receives rewards per block from the fee collector, and `total_accumulated_rewards` are total rewards accumulated in the current epoch
- Accumulate rewards proportionally for each active fp excluding those who does not vote for the target height

#### Per Epoch Boundary:

- Sum each FP's rewards across all processed blocks in the epoch and transfer them to Babylon Genesis
- Clear temporary rewards tracking for the next epoch

### Pros and Cons

Pros:
- rewards are distributed according to voting participation

Cons:
- New parameters and states are introduced
- One misalignment is the rewards are still distributed without considering block finalization. Supporting it requires more sophisticated design

gitferry:
Closes #277. This pr follows a simple approach to improving rewards dist:
- avoid recording fp rewards per block
- distribute fp rewards at rewards dist boundary and counts weighted votes based on its voting power at different heights across a ranged blocks


contracts/btc-finality/src/finality.rs
```diff
@@ -894,52 +894,80 @@ pub fn query_fps_by_total_active_sats(
...
+
+    // Get current balance of the finality contract (total rewards to distribute)
+    let current_balance = deps
         .querier
         .query_balance(env.contract.address.clone(), cfg.denom)?
-        .amount
-        .saturating_sub(total_pending_rewards);
-    // Short-circuit if there are no rewards to distribute
-    if rewards_amount.is_zero() {
+        .amount;
```
SebastianElvis, line 912: We probably need to leverage the reward storage ([ref](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/86b9f83a1eb1d63cb955af4130b600689044649e/contracts/btc-finality/src/state/finality.rs#L64-L69)) for getting the reward to be distributed. This was to prevent the contracts from distributing random tokens that contracts receive from random people

gitferry: I don't understand. We have to obtain the rewards from the balance of the finality contract even though it has balance from random reason, no?

SebastianElvis: The balance needs to be fetched from the `TOTAL_PENDING_REWARDS` storage, rather than the balance. This is because `TOTAL_PENDING_REWARDS` is strictly the value that the BSN allocates to the BTC stakers / FPs, and it might not be equal to the contract balance because the contract can receive tokens from anyone.

gitferry: I feel more confused. `TOTAL_PENDING_REWARDS` only keeps track of the rewards to send to Babylon genesis. We still need query the contracts for balance to distribute rewards. See previous impl https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/3b1af6c09d9dfec8705ff99d33b01ec8c32925c0/contracts/btc-finality/src/finality.rs#L915. In fact, we don't need `TOTAL_PENDING_REWARDS` anymore as we do not need it anymore as we do rewards dist per epoch now

gitferry: Removed it in [c5fb579](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/291/commits/c5fb5790d7298dc54821056606641fad9c4f2295)

contracts/btc-finality/src/finality.rs
```diff
@@ -891,52 +899,62 @@ pub fn query_fps_by_total_active_sats(
...
-    if total_voting_power == 0 {
-        return Ok(());
-    }
-    // Get the rewards to distribute (bank balance of the finality contract, minus previously / already distributed rewards
-    // (pending to be sent to Babylon on an epoch boundary))
-    let total_pending_rewards = TOTAL_PENDING_REWARDS.load(deps.storage)?;
+/// Distributes rewards based on accumulated voting weights from the current reward interval.
+/// This approach uses pre-accumulated voting weights that are collected as FPs submit signatures,
+/// making the distribution O(n) instead of O(m*n) for much better gas efficiency.
+pub fn distribute_rewards_in_range(deps: &mut DepsMut, env: &Env) -> Result<(), ContractError> {
```
liuchengxu, line 905: How about `distribute_rewards_for_interval`?

contracts/btc-finality/src/finality.rs
```diff
@@ -488,6 +488,14 @@ pub fn handle_finality_signature(
     // Store the block height this finality provider has signed
     FP_BLOCK_SIGNER.save(deps.storage, &fp_btc_pk_hex, &height)?;
 
+    // Accumulate voting weight for this FP for reward distribution
+    let power_table = get_power_table_at_height(deps.storage, height)?;
+    if let Some(voting_power) = power_table.get(&fp_btc_pk_hex) {
+        ACCUMULATED_VOTING_WEIGHTS.update(deps.storage, &fp_btc_pk_hex, |existing| {
+            Ok::<u128, ContractError>(existing.unwrap_or(0) + (*voting_power as u128))
+        })?;
+    }
```
SebastianElvis, line 497: Getting the entire voting power table is not efficient, with DB read complexity of `O(n)` where `n` is the number of active FPs.

Alternatively, we can revise L399-400 in a way that

- we get voting power of this FP
- If voting power is 0, return error
- use this voting power at L494

contracts/btc-finality/src/finality.rs
```diff
@@ -891,52 +899,62 @@ pub fn query_fps_by_total_active_sats(
...
+    // Distribute rewards proportionally based on accumulated voting weights
+    for (fp_btc_pk_hex, accumulated_weight) in &accumulated_weights {
+        if *accumulated_weight > 0 {
+            // Use checked multiplication to prevent overflow
+            let numerator = current_balance
+                .u128()
+                .checked_mul(*accumulated_weight)
+                .ok_or(ContractError::CalculationOverflow)?;
+            let reward = numerator / total_accumulated_weight;
+            let reward = Uint128::from(reward);
```
SebastianElvis, line 942: Here `/` might not be a secure operation. We could 

- convert `numerator` and `total_accumulated` weight to `Uint128`
- use `div_floor` function under `Uint128` for division

contracts/btc-finality/src/contract.rs
```diff
@@ -290,14 +285,18 @@ fn send_rewards_msg(
             })
         })
         .collect::<StdResult<Vec<_>>>()?;
+
+    // Calculate total rewards from individual rewards
+    let total_rewards: u128 = fp_rewards.iter().map(|r| r.reward.u128()).sum();
```
SebastianElvis, line 290: Is it possible to 

- let `distribute_rewards_in_range` return the `total_rewards`, and
- pass `total_rewards` as a parameter in the `send_rewards_msg` function?

Summing is `O(n)` operation that ideally we can avoid

contracts/btc-finality/src/contract.rs
```diff
@@ -249,18 +245,18 @@ fn handle_end_block(
...
-            let (fp_rewards, wasm_msg) = send_rewards_msg(deps, rewards.u128(), &cfg)?;
+        distribute_rewards_in_range(deps, &env)?;
+
+        // Then send the accumulated rewards to Babylon Genesis via IBC
+        let (fp_rewards, wasm_msg) = send_rewards_msg(deps, &cfg)?;
+        if !fp_rewards.is_empty() {
             res = res.add_message(wasm_msg);
             // Zero out individual rewards
             for reward in fp_rewards {
                 REWARDS.remove(deps.storage, &reward.fp_pubkey_hex);
```
SebastianElvis, line 258:
```suggestion
                REWARDS.clear(deps.storage);

```

contracts/btc-finality/src/finality.rs
```diff
@@ -891,52 +899,62 @@ pub fn query_fps_by_total_active_sats(
...
-        let reward = Uint128::from(reward);
-        // Update the rewards for this FP
-        REWARDS.update(deps.storage, &fp_btc_pk_hex, |r| {
-            Ok::<Uint128, ContractError>(r.unwrap_or_default() + reward)
-        })?;
-        // Compute the total rewards
-        accumulated_rewards += reward;
+
+    // Distribute rewards proportionally based on accumulated voting weights
+    for (fp_btc_pk_hex, accumulated_weight) in &accumulated_weights {
```
SebastianElvis, line 934: `ACCUMULATED_VOTING_WEIGHTS.range` returns an iterator so we might be able to iterate over it directly, rather than reading those from DB and then iterating

contracts/btc-finality/src/finality.rs
```diff
@@ -891,52 +899,62 @@ pub fn query_fps_by_total_active_sats(
...
+
+    if accumulated_weights.is_empty() {
+        return Ok(()); // No accumulated voting weights
+    }
+
+    // Calculate total accumulated voting weight
+    let total_accumulated_weight: u128 = accumulated_weights.iter().map(|(_, weight)| weight).sum();
+    if total_accumulated_weight == 0 {
         return Ok(());
     }
```
SebastianElvis, line 931: Could we accumulate this value in a `TOTAL_ACCUMULATED_WEIGHT` store upon handling finality signatures as well?

contracts/btc-finality/src/finality.rs
gitferry: Good point, we already load voting power table to check voting power in Line 400, we should not load voting power table again

contracts/btc-finality/src/finality.rs
```diff
@@ -488,6 +488,14 @@ pub fn handle_finality_signature(
     // Store the block height this finality provider has signed
     FP_BLOCK_SIGNER.save(deps.storage, &fp_btc_pk_hex, &height)?;
 
+    // Accumulate voting weight for this FP for reward distribution
+    let power_table = get_power_table_at_height(deps.storage, height)?;
+    if let Some(voting_power) = power_table.get(&fp_btc_pk_hex) {
+        ACCUMULATED_VOTING_WEIGHTS.update(deps.storage, &fp_btc_pk_hex, |existing| {
+            Ok::<u128, ContractError>(existing.unwrap_or(0) + (*voting_power as u128))
```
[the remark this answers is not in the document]
gitferry: Good point. Thing is that `ACCUMULATED_VOTING_WEIGHTS` is not indexed by `(height, fp_btc_pk_hex)`. So we don't know for sure whether the vote is accumulated for the same height (relying on dup checks and evidence checks). But I think we should move this chunk to the condition that the evidence does not exit. See [0c2beac](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/291/commits/0c2beac65cfc13118fd2c440c5147b4cc5b4ac2e)

contracts/btc-finality/src/contract.rs
gitferry: Don't think it's possible as in `distribute_rewards_in_range` we also need to iterate the db to do the same thing. One way though, is to revert [c5fb579](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/291/commits/c5fb5790d7298dc54821056606641fad9c4f2295) to re-introduce `TOTAL_PENDING_REWARDS`

gitferry: Actually I think it's better not to re-introduce `TOTAL_PENDING_REWARDS` as we need to maintain an extra db and should keep its value consistent with the fp rewards sum.

The summing complexity is acceptable as the num of active fps are limited by the params

contracts/btc-finality/src/finality.rs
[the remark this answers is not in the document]
liuchengxu: I see, `ACCUMULATED_VOTING_WEIGHTS` is expected to be modified multiple times in that case as the value is accumulated from a range of blocks. Using `existing.unwrap_or(0)` totally makes sense.

contracts/btc-finality/src/finality.rs
```diff
@@ -396,8 +396,14 @@ pub fn handle_finality_signature(
...
-    // Ensure the finality provider has voting power at this height
-    ensure_fp_has_power(deps.storage, height, fp_btc_pk_hex)?;
+    // Get the finality provider's voting power at this height (also ensures they have power)
+    let voting_power = get_fp_power(deps.storage, height, fp_btc_pk_hex)?;
+    if voting_power == 0 {
+        return Err(ContractError::NoVotingPower(
+            fp_btc_pk_hex.to_string(),
+            height,
+        ));
+    }
```
SebastianElvis, line 406: `get_fp_power` will return error if no voting power so I'm not sure if we still want to keep this if check

contracts/btc-finality/src/finality.rs
```diff
@@ -503,6 +509,11 @@ pub fn handle_finality_signature(
         let (msg, ev) = slash_finality_provider(&mut deps, &fp_btc_pk_hex, &evidence)?;
         res = res.add_message(msg);
         res = res.add_event(ev);
+    } else {
+        // Accumulate voting weight for this FP for reward distribution
+        ACCUMULATED_VOTING_WEIGHTS.update(deps.storage, &fp_btc_pk_hex, |existing| {
+            Ok::<u128, ContractError>(existing.unwrap_or(0) + (voting_power as u128))
+        })?;
     }
```
SebastianElvis, line 517: This could be put between L495 and L497, not necessarily in this else statament

contracts/btc-finality/src/finality.rs
```diff
@@ -891,52 +902,61 @@ pub fn query_fps_by_total_active_sats(
...
+    let mut fp_entries = Vec::new();
+
+    // First pass: collect entries and calculate total
+    for item in
+        ACCUMULATED_VOTING_WEIGHTS.range(deps.storage, None, None, cosmwasm_std::Order::Ascending)
+    {
+        let (fp_btc_pk_hex, weight) = item?;
+        total_accumulated_weight += weight;
+        fp_entries.push((fp_btc_pk_hex, weight));
     }
```
SebastianElvis, line 933: Can we wrap this into a function under `state/`? It can return `fp_entries` and `total_accumulated_weight`

contracts/btc-finality/src/finality.rs
```diff
@@ -891,52 +902,61 @@ pub fn query_fps_by_total_active_sats(
...
+        let accumulated_weight = Uint128::from(accumulated_weight);
+        let numerator = current_balance.checked_mul(accumulated_weight)?;
+        let reward = numerator.div_floor((total_accumulated_weight, Uint128::one()));
+
+        if !reward.is_zero() {
+            // Add reward to FP's pending rewards
+            REWARDS.update(deps.storage, &fp_btc_pk_hex, |existing| {
+                Ok::<Uint128, ContractError>(existing.unwrap_or_default() + reward)
+            })?;
+        }
```
SebastianElvis, line 954: This `if` seems not necessary

contracts/btc-finality/src/contract.rs
```diff
@@ -249,18 +245,16 @@ fn handle_end_block(
         res = res.add_events(events);
     }
 
-    // On an epoch boundary, send rewards for distribution to Babylon Genesis
+    // On a reward distribution boundary, send rewards for distribution to Babylon Genesis
     if env.block.height > 0 && env.block.height % cfg.reward_interval == 0 {
-        let rewards = TOTAL_PENDING_REWARDS.load(deps.storage)?;
-        if rewards.u128() > 0 {
-            let (fp_rewards, wasm_msg) = send_rewards_msg(deps, rewards.u128(), &cfg)?;
+        update_rewards_dist_for_interval(deps, &env)?;
```
SebastianElvis, line 250: It seems that within the if statement of L249, we write into `REWARDS`, read `REWARDS` for constructing reward message, and then clear `REWARDS` right after.

Is it possible to simply avoid this `REWARDS` storage i.e.,

- calls `update_rewards_dist_for_interval` which reads voting power table and contract balance, and returns `fp_rewards` as `Vec<RewardInfo>`
- If `fp_rewards` is empty, then return
- calls `send_rewards_msg` which takes `Vec<RewardInfo>` as input, and returns `wasm_msg`
- adds `wasm_msg` to `res`

This gets rid of most of the DB write operations and thus much cheaper/simpler

contracts/btc-finality/src/finality.rs
gitferry: Hmmm, then the fp that submits conflicting votes could receive rewards for the same height twice, no?

contracts/btc-finality/src/finality.rs
gitferry: I feel it's hard to reason `reward` can never be zero, so keeping it just in case

contracts/btc-finality/src/finality.rs
SebastianElvis: The vote over the fork block will not because the function will return [here](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/98711f8199d01e09ee8b94c71dcb2ca2c6f42a59/contracts/btc-finality/src/finality.rs#L449-L450). Is that correct?

contracts/btc-finality/src/contract.rs
gitferry: Very good point. REWARDS store is legacy and not needed in the current approach. Fixed in [869d53d](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/291/commits/869d53daff3f15ed168ff9f5c23e5c5cdbbf632d)

contracts/btc-finality/src/finality.rs
gitferry: Maybe a cleaner way is the add return in the if statement so that we don't need else statement. Fixed in 
a41a675

PR #294, opened 2025-08-05
slashing: align the slashing implementation with Babylon

### LINKED ISSUE #235: EPIC: misalignment in Cascaded Slashing

SebastianElvis:
Resolves #235

This PR aligns the slashing implementation with Babylon. 

The notable change is to remove the process of extracting secret key and verifying public key correspondance. This is not necessary and Babylon does not do that.

This PR also adds reference to Babylon code base.


contracts/btc-finality/src/finality.rs
```diff
@@ -567,37 +568,12 @@ pub fn handle_unjail(
 
 /// `slash_finality_provider` slashes a finality provider with the given evidence including setting
 /// its voting power to zero, extracting its BTC SK, and emitting an event
+/// following https://github.com/babylonlabs-io/babylon/blob/4aa85a8d9bf85771d448cd3026e99962fe0dab8e/x/finality/keeper/msg_server.go#L384-L412 without the logic for propagating the slashing event to other BSNs
```
liuchengxu, line 571: The comment needs to be updated? (extracting its BTC SK is not true)

contracts/babylon/src/contract.rs
```diff
@@ -269,19 +269,21 @@ pub fn execute(
     match msg {
         ExecuteMsg::Slashing { evidence } => {
             // This is an internal routing message from the `btc_finality` contract
+            // following https://github.com/babylonlabs-io/babylon/blob/4aa85a8d9bf85771d448cd3026e99962fe0dab8e/x/finality/keeper/msg_server.go#L384-L413 without the logic for propagating the slashing event to other BSNs
```
liuchengxu, line 272: Could you also remind me why we send the slashing message to the babylon contract and then the staking contract instead of sending it to the staking contract directly?

SebastianElvis: This is because the finality contract relies on Babylon contract for sending IBC packets about slashing to Babylon. The finality contract does not have access to the zone concierge IBC channel.

liuchengxu: There are two purposes here: one is to send the slashing message for the staking contract for processing, the other is to send the slashing message to Babylon Genesis. Currently, the finality sends one slashing message to the finality contract, and then the finality contract sends a message to the staking contract as well as Babylon Genesis. I mean why not send out two messages from the finality contract, one sent to the staking contract directly, one sent to the babylon contract to propagate it to Babylon Genesis. That being said, the path of sending the slashing message to the staking contract can be done more efficiently, right?

SebastianElvis: > I mean why not send out two messages from the finality contract, one sent to the staking contract directly, one sent to the babylon contract to propagate it to Babylon Genesis. That being said, the path of sending the slashing message to the staking contract can be done more efficiently, right?

Yes this is a potential improvement indeed

PR #300, opened 2025-08-05
chore: move jailing to end blocker

### LINKED ISSUE #266: Misalignment in voting power management
The contracts handle voting power table in [compute_active_finality_providers](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/1eb5086b78ce3fa6aa25d32d40e5548c5f3030b1/contracts/btc-finality/src/finality.rs#L632), while babylon handles voting power table in [UpdatePowerDist](https://github.com/babylonlabs-io/babylon/blob/3d58d818e2f4f93b9e3dd1cad74fe76748db15a9/x/finality/keeper/power_dist_change.go#L26). Both happen upon BeginBlocker.

One major difference is that contacts handles staking events during tx executing in the staking contract, while Babylon handles staking events in the BeginBlocker of the next blocker in the finality module. This does not cause misalignment though.

Some misalignments:

1. Refactor naming of finality provider state https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/5fdbcd7297ddc87b7cd4a6cbdf25ca61f17398a7/contracts/btc-staking/src/state/staking.rs#L66, e.g., power -> [total_bonded_sats](https://github.com/babylonlabs-io/babylon/blob/db2b797e916addb6c1bf46ffd443776be9368577/x/finality/types/finality.pb.go#L93). points_per_stake and points_leftover can be removed
2. [jailing](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/1eb5086b78ce3fa6aa25d32d40e5548c5f3030b1/contracts/btc-finality/src/finality.rs#L698) should not be checked in BeginBlocker along with compute of voting power table. Instead, it should be done at EndBlocker after tallying blocks ([ref(https://github.com/babylonlabs-io/babylon/blob/f6894556b8f7b9ff2008ce7ff1267089ce25edca/x/finality/abci.go#L39))

gitferry:
Closes #266 by moving jailing mechanism to end blocker to align with Babylon Genesis https://github.com/babylonlabs-io/babylon/blob/f6894556b8f7b9ff2008ce7ff1267089ce25edca/x/finality/abci.go#L39.

The PR also added unit tests. E2e tests should be added but in a separate pr.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Improvements
 
 - [#300](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/300) jailing: move jailing to end blocker
+- [#299](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/299) chore: have default values for instantiation messages
+- [#298](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/293) chore: add `test_msg_commit_pub_rand_validate_basic` and minor refactorings
 - [#297](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/297) btc-staking: refactor queries
 - [#294](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/294) slashing: align the slashing implementation with Babylon
 - [#293](https://github.com/babylonlabs-io/cosmos-bsn-contracts/pull/293) ibc: clean up IBC handshake process
--- a/contracts/babylon/src/contract.rs
+++ b/contracts/babylon/src/contract.rs
@@ -7,7 +7,7 @@ use crate::state::config::{Config, CONFIG, DEFAULT_IBC_PACKET_TIMEOUT_DAYS};
 use crate::state::consumer_header_chain::CONSUMER_HEIGHT_LAST;
 use babylon_apis::{btc_staking_api, finality_api, to_bech32_addr, to_module_canonical_addr};
 use cosmwasm_std::{
-    to_json_binary, to_json_string, Addr, Binary, Decimal, Deps, DepsMut, Empty, Env, MessageInfo,
+    to_json_binary, to_json_string, Addr, Decimal, Deps, DepsMut, Empty, Env, MessageInfo,
     QueryResponse, Reply, Response, SubMsg, SubMsgResponse, WasmMsg,
 };
 use cw2::set_contract_version;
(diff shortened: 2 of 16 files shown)


contracts/btc-finality/src/liveness.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        let fp1 = "fp1";
+
+        // Set up power table
+        let mut power_table = HashMap::new();
+        power_table.insert(fp1.to_string(), 1000u64);
+        set_voting_power_table(&mut deps.storage, 100, power_table).unwrap();
+
+        // FP1 has never signed (no entry in FP_BLOCK_SIGNER)
+        // but has a recent start height (within missed_blocks_window)
```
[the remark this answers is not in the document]
gitferry: false negative

contracts/btc-finality/src/liveness.rs
```diff
@@ -0,0 +1,181 @@
...
+                    fp_btc_pk_hex,
+                    |jailed: Option<u64>| match jailed {
+                        Some(jail_time) => Ok::<_, ContractError>(jail_time),
+                        None => Ok(env.block.time.seconds() + cfg.jail_duration),
+                    },
+                )?;
+                Ok(())
+            }
+        }
+    })?;
```
liuchengxu, line 37: Just realize that if it's weird to use `try_for_each()` if no error is returned, we can simply use `for_each()` or a plain loop like the following:

```Rust
let window_start = env.block.height.saturating_sub(cfg.missed_blocks_window);
let now = env.block.time.seconds();

for (fp_btc_pk_hex, _) in fp_power_table {
    let last = get_last_signed_height(deps.storage, fp_btc_pk_hex)?;
    let inactive = match last {
        Some(h) => h <= window_start,
        None => true, // never signed → treat as inactive
    };

    if inactive {
        // Jail if not already jailed; keep the original jail time if present.
        JAIL.update(deps.storage, fp_btc_pk_hex, |existing| {
            Ok::<_, ContractError>(existing.unwrap_or(now + cfg.jail_duration))
        })?;
    }
}
```

It's more readable this way.

contracts/btc-finality/src/liveness.rs
```diff
@@ -0,0 +1,181 @@
+use cosmwasm_std::{DepsMut, Env};
+
+use crate::{
+    error::ContractError,
+    state::config::Config,
+    state::finality::{get_last_signed_height, get_power_table_at_height, JAIL},
+};
```
SebastianElvis, line 7:
```suggestion
use cosmwasm_std::{DepsMut, Env};
use crate::{
    error::ContractError,
    state::config::Config,
    state::finality::{get_last_signed_height, get_power_table_at_height, JAIL},
};
```

contracts/btc-finality/src/liveness.rs
```diff
@@ -0,0 +1,181 @@
...
+                // FP is inactive for at least missed_blocks_window, jail! (if not already jailed)
+                JAIL.update(
+                    deps.storage,
+                    fp_btc_pk_hex,
+                    |jailed: Option<u64>| match jailed {
+                        Some(jail_time) => Ok::<_, ContractError>(jail_time),
+                        None => Ok(env.block.time.seconds() + cfg.jail_duration),
+                    },
+                )?;
+                Ok(())
```
SebastianElvis, line 34: Since we are doing this, should we distinguish between `h <= env.block.height.saturating_sub(cfg.missed_blocks_window)` and the case that `last_sign_height == None`? The former case FP needs to be jailed, but the latter case it means DB is corrupted

SebastianElvis (approved):
Great work :+1:

contracts/btc-finality/src/liveness.rs
gitferry: Great point. I only copied the code and agreed that we should refactor it

PR #305, opened 2025-08-07
feat: add finality_activation_height to params

### LINKED ISSUE #284: finality: introduce finality activation height feature
Currently there is an [activated height](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/a2dfe6c8d3e7bdb4ebf5fbd3fb79074dbb9c6e61/contracts/btc-staking/src/state/staking.rs#L40-L41) in BTC staking contract. It is defined as "the height upon which there exists 1 FP with positive voting power". However, this is legacy code and is no longer precise because the voting power table is now in BTC finality contract.

In addition, we are missing a config value `activation_height` that can be set upon instantiation in BTC finality contract. Babylon has this feature.

- [ ] move activated height from BTC staking contract to BTC finaity contract
- [ ] introduce activation height in instantiation/config of BTC finality contract

gitferry:
Resolve #284

### DIFF
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -1,5 +1,3 @@
-use std::cmp::max;
-
 use crate::error::ContractError;
 use crate::finality::{
     handle_finality_signature, handle_public_randomness_commit, handle_rewards_distribution,
@@ -15,6 +13,7 @@ use crate::state::config::{
 };
 use crate::state::finality::get_btc_staking_activated_height;
 use crate::{finality, queries, state};
+use std::cmp::max;
 
 #[cfg(test)]
 use crate::state::finality::ACCUMULATED_VOTING_WEIGHTS;
--- a/contracts/btc-finality/src/contract.rs
+++ b/contracts/btc-finality/src/contract.rs
@@ -1,3 +1,5 @@
+use std::cmp::max;
+
 use crate::error::ContractError;
 use crate::finality::{
     handle_finality_signature, handle_public_randomness_commit, handle_rewards_distribution,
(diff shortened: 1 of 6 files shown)


contracts/btc-finality/src/contract.rs
```diff
@@ -240,26 +244,26 @@ fn handle_end_block(
...
-    let activated_height = activated_height.unwrap();
-
     // Index the current block
     let ev = finality::index_block(deps, env.block.height, &hex::decode(app_hash_hex)?)?;
     res = res.add_event(ev);
 
     // Tally all non-finalised blocks
-    let events = finality::tally_blocks(deps, &env, activated_height)?;
+    let events = finality::tally_blocks(deps, &env, activated_height.unwrap())?;
     res = res.add_events(events);
```
SebastianElvis, line 267: Here tallying should start since `max(activated_height, cfg.finality_activation_height)`

gitferry: There are still a few mislignments in tally_blocks but out of the scope of this pr. Tracked in https://github.com/babylonlabs-io/cosmos-bsn-contracts/issues/306. Low priority though

SebastianElvis: Also the parameter name in `tally_blocks` could probably be renamed. Currently it's called `activated_height` which might not be precise anymore. Minor stuff though

On the misalignment, I think the two versions are equivalent, and the version in contract is more clean imo actually

gitferry: yeah, this is because we don't want to load activated_height again but passing it as input of `tally_blocks`. We will figure out something when addressing #306

contracts/btc-finality/src/contract.rs
```diff
@@ -1,3 +1,5 @@
+use std::cmp::max;
+
 use crate::error::ContractError;
```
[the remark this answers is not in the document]
liuchengxu: There are long-overdue tasks in #166 that may help with this situation. I can have a look one finishing what I'm currently working on.

PR #308, opened 2025-08-07
chore: change binary to hex string in delegation query

gitferry:
Some important fields in delegation query is not readable as Binary type. This PR changed them to hex string

### DIFF
--- a/contracts/btc-staking/src/state/staking.rs
+++ b/contracts/btc-staking/src/state/staking.rs
@@ -293,7 +293,7 @@ impl From<btc_staking_api::BtcUndelegationInfo> for BtcUndelegationInfo {
 
         BtcUndelegationInfo {
             unbonding_tx: hex::decode(&undelegation_info.unbonding_tx_hex)
-                .expect("valid hex")
+                .expect("failed to decode unbonding_tx_hex from hex")
                 .into(),
             delegator_unbonding_info,
             covenant_unbonding_sig_list: undelegation_info


packages/apis/src/btc_staking_api.rs
```diff
@@ -180,10 +180,10 @@ pub struct ActiveBtcDelegation {
...
     /// Total BTC stakes in this delegation, quantified in satoshi
     pub total_sat: u64,
-    /// Staking tx.
-    pub staking_tx: Binary,
-    /// Slashing tx.
-    pub slashing_tx: Binary,
+    /// Staking tx as hex string.
+    pub staking_tx_hex: String,
+    /// Slashing tx as hex string.
+    pub slashing_tx_hex: String,
```
SebastianElvis, line 186: I don't think this struct should be changed due to the API because it is mostly an internal DB object.

Instead I think we should define dedicated structs for APIs

SebastianElvis: It is an intermediate struct converted from the [IBC packet](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/b298a4d92e8301167eea8d7093d313a691107886/contracts/babylon/src/ibc.rs#L185-L213) (which is part of IBC module's DB). The object corresponding to this struct is used for a series of verifications [here](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/b298a4d92e8301167eea8d7093d313a691107886/contracts/btc-staking/src/staking.rs#L24-L34).

So I think a proper way of fixing this is to create a new struct `BTCDelegationResponse` that can be converted from and to the [BtcDelegation](https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/b298a4d92e8301167eea8d7093d313a691107886/contracts/btc-staking/src/state/staking.rs#L71-L110) DB struct

PR #313, opened 2025-08-08
chore: introduce max blocks to finalize in tallying blocks

### LINKED ISSUE #306: Misalignment in Tally_blocks
1. index_block should be part of tally_blocks, [ref](https://github.com/babylonlabs-io/babylon/blob/f6894556b8f7b9ff2008ce7ff1267089ce25edca/x/finality/keeper/tallying.go#L37)
2. we should introduce a limit to the number of blocks to finalize so that the block processing time is capped https://github.com/babylonlabs-io/babylon/blob/f6894556b8f7b9ff2008ce7ff1267089ce25edca/x/finality/abci.go#L28

gitferry:
Closes #306

### DIFF
--- a/contracts/btc-finality/src/finality.rs
+++ b/contracts/btc-finality/src/finality.rs
@@ -29,10 +29,11 @@ use std::collections::{HashMap, HashSet};
 // or performance degradation caused by excessive future commitments.
 const MAX_PUB_RAND_COMMIT_OFFSET: u64 = 160_000;
 
-// Setting max amount of finalized blocks per EndBlock to 10_000 to cap processing time,
+// Setting max amount of finalized blocks per EndBlock to 1_000 to cap processing time,
 // mirroring Babylon's `MaxFinalizedRewardedBlocksPerEndBlock`.
 // https://github.com/babylonlabs-io/babylon/blob/53d1a8e211f5c9d8b369397bde1f6cf05c7038ad/x/finality/types/constants.go#L7
-pub const MAX_FINALIZED_REWARDED_BLOCKS_PER_END_BLOCK: u64 = 10_000;
+// Setting a smaller value here because the cost in CosmWasm is higher than in Go module.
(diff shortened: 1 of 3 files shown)


contracts/btc-finality/src/finality.rs
```diff
@@ -29,6 +29,11 @@ use std::collections::{HashMap, HashSet};
 // or performance degradation caused by excessive future commitments.
 const MAX_PUB_RAND_COMMIT_OFFSET: u64 = 160_000;
 
+// Setting max amount of finalized blocks per EndBlock to 10_000 to cap processing time,
+// mirroring Babylon's `MaxFinalizedRewardedBlocksPerEndBlock`.
+// https://github.com/babylonlabs-io/babylon/blob/53d1a8e211f5c9d8b369397bde1f6cf05c7038ad/x/finality/types/constants.go#L7
+pub const MAX_FINALIZED_REWARDED_BLOCKS_PER_END_BLOCK: u64 = 10_000;
```
SebastianElvis, line 35: We probably want a smaller value since CW operation is more expensive. Maybe 1000?

contracts/btc-finality/src/finality.rs
```diff
@@ -458,11 +463,16 @@ pub fn index_block(
 pub fn tally_blocks(
     deps: &mut DepsMut,
     env: &Env,
-    activated_height: u64,
+    start_height: u64,
+    max_blocks_to_finalize: u64,
```
SebastianElvis, line 467:
```suggestion
```

can this function access the `MAX_FINALIZED_REWARDED_BLOCKS_PER_END_BLOCK` const directly?

SebastianElvis (approved):

contracts/btc-finality/src/finality.rs
gitferry: we can but this is to align with babylon's implementation. No strong opinion though

contracts/btc-finality/src/finality.rs
gitferry: Sounds good

PR #331, opened 2025-08-13
refactor: introduce `MigrationTester`

liuchengxu:
Created a reusable `MigrationTester` utility to refactor migration tests across all CosmWasm contracts to eliminate code duplication and improve maintainability.
 
 The following `setup()` is redundant and removed in btc-light-client as what `setup()` does is already done in the instantiate function, otherwise, `run_migration_basics()` exactly mirrors the original logic. Therefore, the `MigrationTester` preserves all original test coverage and validation logic.
 
 https://github.com/babylonlabs-io/cosmos-bsn-contracts/blob/8729b1738e66b5f5c212b8913d44e4c8b3663085/contracts/btc-light-client/src/contract.rs#L680
 
 In the future, we can simply extend/modify `MigrationTester` if the standard migration test flow changes. If there is any heavily diverged migration test scenario in some contract, we can just write the specific migration test within the contract.


packages/test-utils/src/migration.rs
```diff
@@ -0,0 +1,303 @@
...
+//!
+//! This module provides a test builder pattern for testing contract migrations
+//! that can be reused across all contracts to avoid code duplication.
+
+use cosmwasm_std::testing::{message_info, mock_dependencies, mock_env};
+use cosmwasm_std::{attr, DepsMut, Env, MessageInfo, Response};
+use cw2::{get_contract_version, set_contract_version};
+use std::fmt::Debug;
+
+/// Type alias for contract migration functions
```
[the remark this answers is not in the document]
liuchengxu: Too verbose, it should be understandable given the context of Migration function.

contracts/btc-finality/src/contract.rs
```diff
@@ -505,73 +506,16 @@ pub(crate) mod tests {
...
+        MigrationTester::new(CONTRACT_NAME, CONTRACT_VERSION).test_migration_basics(
+            migrate,
+            instantiate,
+            crate::msg::MigrateMsg {},
+            InstantiateMsg::default(),
+            |err| match err {
+                ContractError::InvalidContractName { expected, actual } => Some((expected, actual)),
+                _ => None,
+            },
+        );
```
SebastianElvis, line 519: Let's call `test_basic_migration`, `test_after_instantiate`, `test_wrong_contract` given that they are taking different parameters? Same question for other tests

packages/test-utils/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(not(target_arch = "wasm32"))]
```
SebastianElvis, line 1: Why do we need this annotation?

packages/test-utils/src/migration.rs
SebastianElvis, line 1: Super cool!

SebastianElvis (approved):
Great work :+1: Learned something new in Rust

packages/test-utils/src/lib.rs
liuchengxu: Otherwise `cosmwasm::testing` is leaked into the contract compilation, see this failure https://github.com/babylonlabs-io/cosmos-bsn-contracts/actions/runs/16940450022/job/48007836117.

Ideally, the test-utils is not supposed to be compiled in the WASM compilation, but :shrug:

contracts/btc-finality/src/contract.rs
liuchengxu: Not sure I follow your question. Are you saying you want to inline the function `test_migration_basics()` within each contract?

`test_migration_basics()` is a standard migration test suite used to contain all common test scenarios. With this unified API, we can easily extend the standard migration tests in the future.

SebastianElvis: Yeah that was what I mean. Okay then

PR #343, opened 2025-08-22
finality: parameterise max_pub_rand_commit_offset

SebastianElvis:
This PR 

- replaces `MAX_PUB_RAND_COMMIT_OFFSET` const with a conig parameter, passed thru instantiation. This is because the reasonable value for it depends on the block time.
- gives a bigger default value since most Cosmos chains use much shorter block time than Babylon

### DIFF
--- a/contracts/btc-finality/schema/btc-finality.json
+++ b/contracts/btc-finality/schema/btc-finality.json
@@ -37,6 +37,14 @@
         "format": "uint32",
         "minimum": 0.0
       },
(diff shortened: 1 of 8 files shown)


SebastianElvis:
Added a migration handler in the last commit, because this is a breaking change in finality contract's `Config`.

gitferry (approved):
Nice work!

SebastianElvis:
Pushed another commit for updating migration e2e tests

PR #345, opened 2025-08-28
fix: jailing issue in liveness check

gusin13:


gitferry (approved):
Nice fix

gusin13:
I need to merge the PR, in a bit rush will resolve any PR comments separately

PR #346, opened 2025-09-01
chore: LastFinalizedHeight query

gusin13:


contracts/btc-finality/src/migrations/next_height_fix.rs
```diff
@@ -0,0 +1,194 @@
...
+/// causing all subsequent tallying operations to fail with FinalisedBlockWithFinalityProviderSet error.
+///
+/// The fix is very specific and conservative:
+/// 1. ONLY fixes if NEXT_HEIGHT is exactly 342533
+/// 2. ONLY fixes if block 342533 is actually finalized
+/// 3. Sets NEXT_HEIGHT to 342534 to resume finalization
+///
+/// This targeted fix will immediately resume finalization for the 600,000+ blocks that have been stuck
+/// since this specific corruption occurred.
+pub fn fix_next_height_corruption(deps: DepsMut) -> Result<(), ContractError> {
```
gusin13, line 17: this is a manual fix should not be merged, we need to remove this

contracts/btc-finality/src/tallying.rs
```diff
@@ -47,6 +47,11 @@ pub(crate) fn tally_blocks_with_max_finalized_blocks(
     let next_height = NEXT_HEIGHT.may_load(deps.storage)?.unwrap_or(0);
     let start_height = max(start_height, next_height);
 
+    deps.api.debug(&format!(
```
gusin13, line 50: we need these debug stuff right now, will remove later when incident resolved

PR #355, opened 2025-09-10
chore: adds batch fp power query

### LINKED ISSUE #354: Batch call for FinalityProviderPower
Finality provider uses `FinalityProviderPower` to determine if a given FP has voting power at a specific block height. This is used to filter out blocks in a batch for which the FP doesn't have voting power to avoid invalid submission. 

For babylon FP the batch submission size is 1000, meaning that per batch, the FP would make 1000 RPC calls. This seems like an unnecessary network and RPC load that could be mitigated by a batch operation.

This issue proposes a  `FinalityProviderPowers` with the following request:
```go
type FinalityProviderPowerQuery struct {
	BtcPkHex string `json:"btc_pk_hex"`
	Heights  []uint64 `json:"heights"`
}

type FinalityProviderPowersResult struct {
    Powers   []FinalityProviderPowerResult `json:"powers,omitempty"`
}

type FinalityProviderPowerResult struct {
    Height      uint64 `json:"height,omitempty"`
    VotingPower uint64 `json:"voting_power,omitempty"`
}
```
Request Limits

Maximum total height queries per request: 1000 (proposed number, can be higher/lower)
Missing voting power: Returns 0 for heights where the FP has no voting power
Response order: Results array maintains the same order as input queries

Lazar955:
closes #354

### DIFF
--- a/contracts/btc-finality/src/queries.rs
+++ b/contracts/btc-finality/src/queries.rs
@@ -1,7 +1,7 @@
 use crate::error::ContractError;
 use crate::msg::{
     ActiveFinalityProvidersResponse, BlocksResponse, EvidenceResponse,
-    FinalityProviderPowerResponse, FinalityProviderPowerBatchResponse, FinalitySignatureResponse,
+    FinalityProviderPowerBatchResponse, FinalityProviderPowerResponse, FinalitySignatureResponse,
     JailedFinalityProvider, JailedFinalityProvidersResponse, SigningInfoResponse, VotesResponse,
 };
 use crate::state::finality::{
@@ -134,7 +134,7 @@ pub fn finality_provider_power_batch(
     }
 
     let mut powers = Vec::with_capacity(heights.len());
-    
+
     for height in heights {
         let power_table = get_power_table_at_height(deps.storage, height)?;
         let power = power_table.get(&btc_pk_hex).copied().unwrap_or(0);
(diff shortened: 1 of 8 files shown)


contracts/btc-finality/src/queries.rs
```diff
@@ -1,22 +1,25 @@
...
-use cosmwasm_std::{Deps, StdResult};
+use cosmwasm_std::{Deps, StdError, StdResult};
 use cw_storage_plus::Bound;
 
 // Settings for pagination
 const MAX_LIMIT: u32 = 30;
 const DEFAULT_LIMIT: u32 = 10;
 
+// Maximum number of heights allowed in batch queries
+const MAX_HEIGHTS_LIMIT: usize = 100;
```
SebastianElvis, line 21:
```suggestion
// Maximum number of heights allowed in batch queries
const MAX_HEIGHTS_LIMIT: usize = 30;
```

to align with other limits. This is due to costs in CW operation

SebastianElvis (approved):

gusin13 (approved):

contracts/btc-finality/src/queries.rs
Lazar955: Is 30 the upper limit? could we squeeze 50 😅

SebastianElvis: we can try but not 100% sure

Lazar955: image

Managed to return 100 from the contract. Is it okay to leave 100 in then

PR #363, opened 2025-09-12
fix: add proof of work validation in btc headers

gusin13:


gusin13:
closing for now, as discussed on slack

we have trust assumption that contract will accept all btc headers from babylon genesis w/o verification due to complexity and gas issues around cryptography verification in contract side cc @SebastianElvis

